PR #2, opened 2015-04-14
Unused Application::container property

iwankgb:
The ``container`` property in ``Application`` class does not seem to be used.


taylorotwell:
Can send this to 5.0 branch.

PR #6, opened 2015-04-14
Application::getMethod() and Application::getPathInfo()

iwankgb:
There is no reason to rely on ``getMethod()`` and ``getPathInfo()`` methods from ``Laravel\Lumen\Application`` as respective methods of ``Symfony\Component\HttpFoundation\Request`` provides very same functionality.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -1052,13 +1052,11 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
      */
     public function dispatch($request = null)
     {
-        if ($request) {
-            $method = $request->getMethod();
-            $pathInfo = $request->getPathInfo();
-        } else {
-            $method = $this->getMethod();
-            $pathInfo = $this->getPathInfo();
+        if ($request === null) {
+            $request = $this->make('request');
         }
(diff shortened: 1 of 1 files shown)


mlukaszewski:
:+1:

GrahamCampbell:
This is intentional. Luman doesn't create a symfony request object if it doesn't need to in order to save time.

iwankgb:
@GrahamCampbell - what about ``getMethod()`` allowing to use non-existent method names? I don't think it's a very good idea to be honest.
Regarding not creating Symfony object for performance reasons - as far as I understand ``Application::dispatch()`` is intended to be called from ``Application::handle()`` (and will never be null in this case) or ``Application::run()`` (and might be null in this case). Wouldn't it make sense to drop implementation of ``HttpKernelInterface`` in this case? It does not seem to be used anywhere in the application.

iwankgb:
An update: ``run()`` method seems to be used when handling HTTP request but ``handle()`` method is being used in tests - so code used to handle HTTP requests is not really being tested.

PR #7, opened 2015-04-14
Application::register() $provider parameter

iwankgb:
According to phpdoc for ``Illuminate\Contracts\Foundation\Application::register()`` and ``Laravel\Lumen\Application::register()`` it must be possible to pass instance of ``Illuminate\Support\ServiceProvider`` as first parameter. This pull request makes it really an option.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -209,19 +209,20 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
      */
     public function register($provider, $options = array(), $force = false)
     {
-        if (array_key_exists($provider, $this->loadedProviders)) {
+        if (!$provider instanceof \Illuminate\Support\ServiceProvider) {
+            $provider = new $provider($this);
+        }
+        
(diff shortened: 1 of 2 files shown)


src/Application.php
```diff
@@ -209,19 +209,20 @@ public function registerConfiguredProviders()
      */
     public function register($provider, $options = array(), $force = false)
     {
-        if (array_key_exists($provider, $this->loadedProviders)) {
+        if (!$provider instanceof \Illuminate\Support\ServiceProvider) {
```
GrahamCampbell: this needs importing

src/Application.php
```diff
@@ -209,19 +209,20 @@ public function registerConfiguredProviders()
...
+            $provider = new $provider($this);
+        }
+        
+        $class = new \ReflectionClass($provider);
```
GrahamCampbell: import please

PR #8, opened 2015-04-14
PSR-2 formatting

iwankgb:
Bunch of formatting fixes that make source code as PSR-2 friendly as possible.


GrahamCampbell:
Most of these aren't PSR-2 changes. Please revert all changes that are marked as "SHOULD" in the PSR-2 spec. Also, some changes are even totally unrelated.

PR #9, opened 2015-04-14
PSR-2 in tests

iwankgb:
Following changes have been applied:
* each class lives in its own file (PSR-1);
* each class has a namespace defined (PSR-1);
* lines longer then 120 lines have been broken into multiple lines (PSR-2);
* ExampleTest class is renamed to FullApplicationTest (PSR-4 - yes, I know it is not part of PSR-2 but it's close enough to me).

Tests had to be modified a bit and TestController file moved to another location to accommodate changes listed above.

Some additional issues have been identiefied:
* ``Laravel\Lumen\Application::handle()`` method is beind used in tests while in real world ``Laravel\Lumen\Application::run()`` [will be used](https://github.com/laravel/lumen/blob/master/public/index.php#L28);

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -54,7 +54,8 @@
     },
     "autoload": {
         "psr-4": {
-            "Laravel\\Lumen\\": "src/"
+            "Laravel\\Lumen\\": "src/",
+            "Laravel\\Lumen\\Tests\\": "tests/"
         },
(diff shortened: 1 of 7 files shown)


taylorotwell:
See comment on previous PR.

PR #10, opened 2015-04-14
PSR-2 in production code

iwankgb:
Following types of changes have been applied to observe PSR-2:
* lines longer that 120 lines are broken into many lines (yes, it is a soft limit);
* method calls incorrectly broken into more then 1 line are fixed;
* some spacing issues are fixed too.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -370,7 +370,11 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
         });
 
         $this->singleton('auth.password', function () {
-            return $this->loadComponent('auth', 'Illuminate\Auth\Passwords\PasswordResetServiceProvider', 'auth.password');
+            return $this->loadComponent(
+                'auth',
+                'Illuminate\Auth\Passwords\PasswordResetServiceProvider',
+                'auth.password'
+            );
         });
     }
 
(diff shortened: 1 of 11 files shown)


taylorotwell:
I only tend to accept PSR-2 fixes that can be automated by php-cs-fixer as its more scalable going forward. Thanks

GrahamCampbell:
Also, these changes aren't PSR-2. PSR-2 does NOT stipulate a hard line length.

src/Application.php
```diff
@@ -370,7 +370,11 @@ protected function registerAuthBindings()
         });
 
         $this->singleton('auth.password', function () {
-            return $this->loadComponent('auth', 'Illuminate\Auth\Passwords\PasswordResetServiceProvider', 'auth.password');
+            return $this->loadComponent(
```
MacDada: @GrahamCampbell You can argue that `PSR-2` does not require this…

src/Providers/CacheServiceProvider.php
```diff
@@ -3,7 +3,8 @@
 use Illuminate\Cache\Console\ClearCommand;
 use Illuminate\Cache\CacheServiceProvider as BaseProvider;
 
-class CacheServiceProvider extends BaseProvider {
+class CacheServiceProvider extends BaseProvider
```
MacDada: @taylorotwell @GrahamCampbell …but this clearly is a violation.

PR #11, opened 2015-04-14
Add the awesome laravel key generate command

abstractFlo:
Please add this PR for enable the awesome command for generating app keys.

I have changed a little bit from the laravel version. No need for app.php in config folder. Overwrite APP_KEY in .env file easy

### DIFF
--- a/src/Console/Commands/KeyGenerateCommmand.php
+++ b/src/Console/Commands/KeyGenerateCommmand.php
@@ -29,7 +29,7 @@ class KeyGenerateCommand extends Command {
 	{
 		$key = $this->getRandomKey();
 
-		if($this->confirm('You have uncomment Dotenv::load in boostrap/app.php? [yes|no]')) {
+		if($this->confirm('Have you uncommented Dotenv::load in bootstrap/app.php? [yes|no]')) {
 			if ($this->option('show')) {
 				return $this->line('<comment>'.$key.'</comment>');
 			}
--- a/src/Console/Commands/KeyGenerateCommmand.php
+++ b/src/Console/Commands/KeyGenerateCommmand.php
@@ -1,0 +1,75 @@
+<?php namespace Laravel\Lumen\Console\Commands;
+
+use Illuminate\Support\Str;
+use Illuminate\Console\Command;
+use Symfony\Component\Console\Input\InputOption;
+
+class KeyGenerateCommand extends Command {
+
+	/**
+	 * The console command name.
+	 *
+	 * @var string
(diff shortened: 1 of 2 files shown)


jbrooksuk:
:+1: I was just about to do the same thing.

GrahamCampbell:
You must use spaces for indentation.

GrahamCampbell:
Also, I'm :-1: on adding anything more to Lumen. It's supposed to be really light and basic.

abstractFlo:
GrahamCampbell i can't understand your problem. Ist only a simple helper for generating the Key. I think all people love this. 

no key you have to come up with and has therefore not always temptation to use the same.

GrahamCampbell:
@abstractFlo My point is that if people want this, they should probably be using Laravel, not Lumen.

abstractFlo:
What? Thats not the point. It's a simple command. Not more, Lumen is awesome and only for this fu... Command i must use laravel? Not really

GrahamCampbell:
@abstractFlo It's not just this command. We have to draw a line somewhere, or everything from Laravel will end up in Lumen because it's "cool".

abstractFlo:
that's the point is not, it's still just a matter of simple, quickly and safely to generate a key app. no more and no less

but if you mean that everyone who wants to use this command lumen, not to use, then it is so.

I can not understand

GrahamCampbell:
My issue is with us registering a command that is only useful once.

abstractFlo:
yes, and thats the point. Only useful once, for secure session, cookies, crypt.....

jbrooksuk:
What if this was a package?

GrahamCampbell:
It might be cool to get this integrated into the illuminate installer perhapse?

abstractFlo:
@jbrooksuk thats not my opinion. it's a small command. i think this is a base feature of a secure application. Helps people to get secure keys....

jbrooksuk:
@GrahamCampbell I like that idea.

@abstractFlo the problem with adding another command (just like #23) is that the IOC has to register another class, it adds overhead.

abstractFlo:
It's your choice, i think there is no more overhead with this command.

jbrooksuk:
But of course there is overhead. It's more to be loaded in. Whilst they may be light classes, if you keep adding them, then it's not light anymore :)

tommymarshall:
Maybe on `post-install` grab the contents of a randomly generated key from a service/app that is also running Lumen? (key.lumen.laravel.com) How meta would that be ;)

PR #12, opened 2015-04-14
Add PSR-2 compatibility using php-cs-fixer.

crynobone:
`$ ./vendor/bin/php-cs-fixer fix`, could also utilize @GrahamCampbell StyleCI.io

Signed-off-by: crynobone <crynobone@gmail.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /lumen.log
 /vendor
 composer.lock
+.php_cs.cache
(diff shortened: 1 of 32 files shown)


taylorotwell:
I prefer to just use `php-cs-fixer fix --level=psr2` I think.

GrahamCampbell:
@crynobone We will be using the rules from the the laravel preset on styleci.

PR #15, opened 2015-04-15
fix error 404 if not run Lumen in root

ptcong:


lukasgeiter:
I noticed this as well, but I'm wondering if this fix covers every case. L5 uses [this method of the symfony request class](https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/HttpFoundation/Request.php#L1792-L1851) to determine the base url. Which is *a lot* more than just `dirname($_SERVER['SCRIPT_NAME'])`

mayoz:
I think, it should be used only Symfony Request.

ptcong:
Both are same, and Lumen is a micro framework and should not make it slower

PR #20, opened 2015-04-15
Pagination: Generated urls cauase a 404

KluVerKamp:
The urls generated in the pagination palette causes a 404 error.
The main cause is the trailing / at the end of the current uri. Tried to take it away and works fine.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,34 +16,34 @@
     ],
     "require": {
         "php": ">=5.4.0",
-        "illuminate/auth": "5.0.*",
-        "illuminate/bus": "5.0.*",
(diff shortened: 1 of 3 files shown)


taylorotwell:
This PR includes a lot of other changes. Can you make another fresh one?

KluVerKamp:
My purpose was to report the issue related to the pagination described in PR description.
I couldn't make an empty PR, sorry for the fuss

KluVerKamp:
it seems like the Fast-Router doesn't accept trailing slash at the end.
When running on the php server this problem happens. But when running on apache/nginx, the webserver redirects in order to remove the trailing slash.

PR #21, opened 2015-04-15
Fix error 404 if not run Lumen in root

ptcong:
`getPathInfo()` give wrong uri if Lumen run in a sub folder, this PR to resolve the problem.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -1324,9 +1324,9 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
     public function getPathInfo()
     {
         $query = isset($_SERVER['QUERY_STRING']) ? $_SERVER['QUERY_STRING'] : '';
-        $folder = dirname($_SERVER['SCRIPT_NAME']).'/';
+        $folder = dirname($_SERVER['SCRIPT_NAME']);
         $uri = $_SERVER['REQUEST_URI'];
-        if (strpos($uri, $folder) === 0) {
+        if ($folder != $uri && strpos($uri, $folder) === 0) {
             $uri = substr($uri, strlen($folder));
         }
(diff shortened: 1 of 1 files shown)


ptcong:
I just submited new PR #29 to cover a missed case

src/Application.php
```diff
@@ -1324,8 +1324,12 @@ protected function getMethod()
     public function getPathInfo()
     {
         $query = isset($_SERVER['QUERY_STRING']) ? $_SERVER['QUERY_STRING'] : '';
-
-        return '/'.ltrim(str_replace('?'.$query, '', $_SERVER['REQUEST_URI']), '/');
+        $folder = dirname($_SERVER['SCRIPT_NAME']);
```
GrahamCampbell: folder?

GrahamCampbell:
Oh, right, I see. You can't run lumen in a subfolder.

src/Application.php
ptcong: Ya, if i run lumen in http://localhost/folder/folder/ then the $folder is **folder/folder/**
If run in root domain, the $folder is **/**

ptcong:
So then will this fixed ?

GrahamCampbell:
It's not going to be fixed. It's not supported.

ptcong:
:-1: people must run lumen in root domain.
What about if i have used laravel in root domain, and want to use lumen in http://domain/api/ ? 
don't tell me that i must setup api.domain for it to use.

GrahamCampbell:
@ptcong That is also true for Laravel.

ptcong:
With laravel and all other framework i have used, i can run in subfolder as well.

GrahamCampbell:
We don't encourage it in Laravel, and we pretend you can't in the docs.

PR #23, opened 2015-04-15
Re-add maintenance mode.

jbrooksuk:
Maintenance mode now works again. The actual [lumen](https://github.com/laravel/lumen) app just needs the middleware adding for `CheckForMaintenanceMode`.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -186,7 +186,7 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
      */
     public function isDownForMaintenance()
     {
-        return false;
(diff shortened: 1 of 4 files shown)


jbrooksuk:
I love that this was partially implemented.

ptcong:
Should re-add `app:name` too ?

jbrooksuk:
Yeah, but in a separate commit.

taylorotwell:
I'm purposefully keeping these types of things out of the repo. Also, something like Envoyer is really the better way to go as even maintenance mode will give you down time.

GrahamCampbell:
This is something you can add to your own application if you want to anyway. :)

jbrooksuk:
That's how it started out for me :)

PR #24, opened 2015-04-15
Fix error when using PHP built-in server

WouterJ:
When using `php -S localhost:1234 server.php` to run Lumen, php_sapi_name is `cli-server` instead of `cli` (at least, on my Windows PC). This PR makes sure both are recognized as CLI and also allows other variants of cli-*.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -1435,7 +1435,7 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
      */
     public function runningInConsole()
     {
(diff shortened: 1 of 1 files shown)


taylorotwell:
runningInConsole is mainly used for determining if you are running Artisan, not the PHP web server.

WouterJ:
@taylorotwell then you should update `Application#storagePath()`, as `getcwd()` should be used instead of `realpath(getcwd().'/../')` when using the built-in web server to run a Lumen application.

PR #25, opened 2015-04-15
Additional method to register a  controller route

ptcong:
I know that purpose of Lumen is performance, and i love it. 
But for me and many people, not only use micro-framework to build restAPI, still need to build a full features web application with backend.
For building routes for backend, i think controller route is very helpful because repeat multiple times with $app->get(), $app->post() seem very stupid :(

Should consider to addition this PR. @Otwell


taylorotwell:
Not going to include this.

PR #27, opened 2015-04-15
[5.0] Add more description to php artisan make command

Arrilot:
Makes more clear what we can do with this command without digging into the source code.


GrahamCampbell:
Duplicate of #13.

Arrilot:
Just noticed it a second ago, sorry :(

PR #28, opened 2015-04-16
Fix Syntax error in config/filesystems.php

camelCaseD:


taylorotwell:
Where is the syntax error?

camelCaseD:
It was causing you to add `']'' on that line, before redoing the white space.

camelCaseD:
For further clarification this I'm using shows what that character was... screenshot_2015-04-19-11-38-45

PR #29, opened 2015-04-16
fix error 404 if not run Lumen in root [2]

ptcong:
Previous PR, i have missing a case that if uri contains a path like sub-folder name


jeteon:
I develop web applications for people so my local URIs are almost always in a sub-directory. This really helped.

ptcong:
Taylor Otwell have rejected the PR

jeteon:
Yeah, I saw. It really helped me out though. I'm sure others who come across the issue will find it helpful as well.

PR #31, opened 2015-04-16
Route groups should be inherit parent group attributes

AlifAmri:
The current situation, if we have group inside another group, the child group is not able to get parent attributes. It is important since we can using namespacing group. 

the code example will explain the rest. 

```php
$app->group( [ 'namespace' => 'App\Http\Controllers' ], function ( $app ){
    $app->get( '/', [
        'uses' => 'SomeController@getIndex'
    ] );
    
    $app->group( [ 'middleware' => 'auth' ], function ( $app ){
        // this will be failed because parent group parameters has been nulled
        $app->get( 'foo', [
            'uses' => 'FooController@getIndex'
        ] );
    });

    $app->group( [ 'middleware' => 'auth.basic' ], function ( $app ){
        $app->get( 'bar', [
            'uses' => 'BarController@getIndex'
        ] );
    } );
});
```

if we print the routes, we will get this
```sh
array:3 [
  "GET/" => array:3 [
    "method" => "GET"
    "uri" => "/"
    "action" => array:1 [
      "uses" => "App\Http\Controllers\SomeController@getIndex"
    ]
  ]
  "GET/foo" => array:3 [
    "method" => "GET"
    "uri" => "/foo"
    "action" => array:2 [
      "uses" => "FooController@getIndex"
      "middleware" => "auth"
    ]
  ]
  "GET/bar" => array:3 [
    "method" => "GET"
    "uri" => "/bar"
    "action" => array:2 [
      "uses" => "BarController@getIndex"
      "middleware" => "auth.basic"
    ]
  ]
]
```

 I hope this PR will fix this

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -782,12 +782,12 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
      */
     public function group(array $attributes, Closure $callback)
     {
-        $previous_attributes   = $this->groupAttributes;
+        $previousAttributes = $this->groupAttributes;
         $this->groupAttributes = $this->mergeParentGroupAttributes($attributes);
 
         call_user_func($callback, $this);
 
-        $this->groupAttributes = $previous_attributes;
+        $this->groupAttributes = $previousAttributes;
     }
 
     /**
@@ -918,13 +918,12 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
     /**
      * Merge parent group attributes.
      *
-     * @param array $attributes
+     * @param  array $attributes
      * @return array
      */
     protected function mergeParentGroupAttributes(array $attributes)
     {
         if (isset($this->groupAttributes)) {
-
             return array_merge($this->groupAttributes, $attributes);
         }
 
--- a/src/Application.php
+++ b/src/Application.php
@@ -782,11 +782,12 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
      */
     public function group(array $attributes, Closure $callback)
     {
-        $this->groupAttributes = $attributes;
+        $previous_attributes   = $this->groupAttributes;
+        $this->groupAttributes = $this->mergeParentGroupAttributes($attributes);
 
         call_user_func($callback, $this);
 
-        $this->groupAttributes = null;
+        $this->groupAttributes = $previous_attributes;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Application.php
```diff
@@ -782,11 +782,12 @@ public function withEloquent()
      */
     public function group(array $attributes, Closure $callback)
     {
-        $this->groupAttributes = $attributes;
+        $previous_attributes   = $this->groupAttributes;
```
GrahamCampbell: You MUST use camelCase, not snake_case.

GrahamCampbell: Also, you must not align the equals like that.

src/Application.php
```diff
@@ -915,6 +916,22 @@ protected function parseAction($action)
     }
 
     /**
+     * Merge parent group attributes.
+     *
+     * @param array $attributes
```
GrahamCampbell: cs

src/Application.php
```diff
@@ -915,6 +916,22 @@ protected function parseAction($action)
...
+     * @return array
+     */
+    protected function mergeParentGroupAttributes(array $attributes)
+    {
+        if (isset($this->groupAttributes)) {
+
```
GrahamCampbell: no blank line plz

GrahamCampbell:
How does this compare performance wise?

AlifAmri:
performance hopefully will be improved since we dont need to redefine namespace attribute on each group

taylorotwell:
You're not just going to be able array_merge the namespace attributes. They have to manually merged since you really want to concatenate strings.

AlifAmri:
@taylorotwell  hm why not? ah my mistake ... 
The real problem is not on namespace attribute but on groupAttributes it self,

let me see you some nasty example

```php
$app->group( [ 'middleware' => 'auth' ], function ( $app ){
    // groupAttributes middleware was set
    $app->group( [ 'namespace' => 'App\Http\Controllers' ], function ( $app ){
        // groupAttributes was set again, and loose groupAttributes middleware
        $app->get( 'foo', [
            'uses' => 'DashboardController@getIndex'
        ] );
        // groupAttributes was nulled
    } );
    
   // where is my groupAttributes ?
   // oh snap!! its being nulled by previous group
    $app->get( '/', function (){
        return response( 'Hello World' );
    } );

} );
```

the first group attributes, 'middleware auth' is not applying into every route inside himself.
Maybe on current release we still no need recrusive group, because its just group namespace and middleware, but maybe you considering to give lumen 'group prefix' that would be awsome :) 

thanks,

PR #32, opened 2015-04-16
Support built-in PHP server

WouterJ:
Second try at fixing errors when using the built-in PHP server + `server.php` router file with Lumen.


WouterJ:
Thanks @taylorotwell.

PR #39, opened 2015-04-19
Fix Log resolver

quetzyg:
Currently, when using any of the `Log` facade methods we will get a fatal error:
```Fatal error: Call to undefined method Illuminate\Support\Facades\Log::error() in vendor/illuminate/support/Facades/Facade.php on line 210```

This is due to the Laravel `Log` facade `getFacadeAccessor()` method returning `log`, while in the `Application.php` we have `Psr\Log\LoggerInterface` instead.

This pull request fixes that.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -526,7 +526,7 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
      */
     protected function registerLogBindings()
     {
-        $this->singleton('log', function () {
+        $this->singleton('Psr\Log\LoggerInterface', function () {
             return new Logger('lumen', [$this->getMonologHandler()]);
         });
     }
@@ -1496,6 +1496,7 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
             'request' => 'Illuminate\Http\Request',
             'Illuminate\Session\SessionManager' => 'session',
             'Illuminate\Contracts\View\Factory' => 'view',
+            'log' => 'Psr\Log\LoggerInterface',
         ];
     }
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
This fix is wrong. It brakes the binding to the interface.

quetzyg:
Well, in that case the fix would probably have to be made in `Illuminate\Support\Facades\Log.php`.

Changing from `return 'log';` to `return 'Psr\Log\LoggerInterface';`

But that would also mess with Laravel, not just Lumen.

GrahamCampbell:
@quetzyg No, both need to be aliased.

quetzyg:
Yeah, you're right @GrahamCampbell. It's definitely be fixed now :)

GrahamCampbell:
That still looks wrong ti me,

quetzyg:
OK, care to elaborate?

taylorotwell:
This is correct.

Someguy123:
@taylorotwell ETA until this is in composer (stable)? `composer update` isn't getting this fix yet

taylorotwell:
I'll tag it tomorrow.

GrahamCampbell:
@Someguy123 You need to set the correct stability then.

taylorotwell:
Graham: we wouldn't tell people to ride dev stability for bug fixes. That
would be bad advice to give.

GrahamCampbell:
> would be bad advice to give.

I totally disagree. Requiring the lumen dependency with dev is a good idea if you want bug fixes early. That's the entire point in dev stability!

GrahamCampbell:
Note that this is not the same as requiring all deps with dev stability - that would normally be a bad idea.

taylorotwell:
I'll talk to you about this elsewhere but please do not give this advice to
people on Laravel GitHub.

PR #44, opened 2015-04-21
Fix basepath for artisan serve

petecoop:
Running from `php artisan serve` in `v5.0.3` causes:

```
Fatal error: Uncaught exception 'UnexpectedValueException' with message 'The stream or file "<redacted>/public/storage/logs/lumen.log" could not be opened: failed to open stream: No such file or directory' 
```

I fixed the issue by removing these lines however I'm sure they are there for a reason. I've not done anything other than install, create a `home.php` view then `return view('home');` `in routes.php`

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -1338,11 +1338,7 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
             return $this->basePath.($path ? '/'.$path : $path);
         }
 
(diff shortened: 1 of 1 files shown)


petecoop:
Is artisan serve not supported?

GrahamCampbell:
No it is supported. This fix just doesn't look correct.

petecoop:
Fair enough - I didn't think this would be the appropriate fix, but found no other way to open up the conversation, no issues?

PR #46, opened 2015-04-22
Add support to use redis cache driver

danwall:
Allow the `redis` cache driver to be able to be used as per the documentation.

Have also added env config options for redis host/port and the cache prefix.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,6 +32,7 @@
         "illuminate/http": "5.0.*",
         "illuminate/pagination": "5.0.*",
         "illuminate/queue": "5.0.*",
+        "illuminate/redis": "5.0.*",
         "illuminate/session": "5.0.*",
         "illuminate/support": "5.0.*",
         "illuminate/translation": "5.0.*",
--- a/config/cache.php
+++ b/config/cache.php
@@ -70,6 +70,6 @@ return [
     |
     */
 
-    'prefix' => 'laravel',
+    'prefix' => env('CACHE_PREFIX', 'laravel'),
 
 ];
--- a/config/database.php
+++ b/config/database.php
@@ -121,8 +121,8 @@ return [
         'cluster' => false,
 
         'default' => [
-            'host'     => '127.0.0.1',
-            'port'     => 6379,
+            'host'     => env('REDIS_HOST', '127.0.0.1'),
+            'port'     => env('REDIS_PORT', 6379),
             'database' => 0,
         ],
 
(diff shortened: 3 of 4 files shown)


GrahamCampbell:
This belongs in your own app. Lumen isn't adding this extra overhead on purpose in the interest of staying light.

GrahamCampbell:
As proved here, registering the redis stuff is pretty simple, so it's not going to take anyone very long to stick this code into their app. :)

danwall:
Would you like me to send a PR to lumen-docs outlining the process? It's quite a bit more complicated than simply changing the .env file and pulling in `predis/predis` as the docs suggest.

Even then it feels half complete without at least having the config changes from this PR merged.

Thoughts?

jaden:
@danwall Updating the docs seems like the right approach. It keeps Lumen fast out of the box, but lets users know how to add it, since more is required than simply installing predis/predis via Composer.

dennisoderwald:
I stood in front of the same puzzle..

chriswonders:
You just need to pull in illuminate/redis to fill the gap.

danwall:
Yeah this has mostly been added back in since this PR was closed and docs have been updated which is great.

https://github.com/laravel/lumen-framework/commit/7aaa53110ddffc1dc143d3e2e8172ab79f87785b
https://github.com/laravel/lumen-framework/commit/d23ae5c904ceb760b2d5433f357fab7f95d7320a

There still appear to be issues when trying to type hint the cache contract in a controller though which is annoying.

Adding something like the below in your AppServiceProvider seems to do the trick.

```php
$this->app->singleton('cache.store', function () {
    $this->app->configure('cache');
    $this->app->register('Illuminate\Cache\CacheServiceProvider');
    return $this->app->make('cache.store');
});
```

dennisoderwald:
thanks @chriswonders!

approots:
@danwall - maybe things have changed, but to inject the cache by type-hinting in a class constructor use:

    use Illuminate\Contracts\Cache\Repository;

    class MyClass {
        public function __construct(Repository $cache) {
        ...

I tried using everything else, but turns out it's mentioned in the docs under cache usage: http://lumen.laravel.com/docs/cache#cache-usage

ctrlaltdylan:
Using Lumen 5.2, like @GrahamCampbell was saying you just need to update your app registration with the `illuminate\redis` provider:

```
$app->register(Illuminate\Redis\RedisServiceProvider::class);
```

PR #47, opened 2015-04-22
Fix basePath for artisan serve command

mattbit:
The bug was introduced in https://github.com/laravel/lumen-framework/commit/e5e4ac6e082c153c5f9581f6236222a2516b6f26.

The PHP built-in server still works fine if executed with the correct document root (e.g. `php -t path_to_my_lumen_app/public -S 127.0.0.1:8000`).

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -1338,7 +1338,7 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
             return $this->basePath.($path ? '/'.$path : $path);
         }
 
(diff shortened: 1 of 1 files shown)


electrickite:
This is only a problem because the working directory is changed in `Laravel\Lumen\Console\Commands\ServeCommand.php`

    public function fire()
    {
        chdir(base_path('public'));

taylorotwell:
I'm just going to pass the base path explicitly like full laravel becuase this is getting annoying.

PR #52, opened 2015-04-25
Add Route Parameter Helper

irazasyed:
Helper to get route parameter. Comes handy in middleware and elsewhere.
Available in Laravel. But was missing in Lumen.


GrahamCampbell:
This should be defined in your own app. Lumen is not trying to be Laravel.

PR #56, opened 2015-04-27
groupAttributes can be used with router group nesting

gloomyzerg:
example:
```php
$app->group(['namespace'=>'App\Http\Controllers', 'prefix'=>'test'], function() use ($app) {

    $app->get('/', function() use ($app) {
        return $app->welcome();
    });

    $app->group(['prefix'=>'some'], function() use ($app) {
        $app->get('/', 'SomeController@index');
    });
});
```

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -91,7 +91,7 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
      *
      * @var array|null
      */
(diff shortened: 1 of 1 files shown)


taylorotwell:
You can't just do a simple merge like that. That won't work for prefixes.

PR #64, opened 2015-04-30
Changed the default cache driver to file.

sunspikes:
Doesn't it make sense to keep the default driver as 'file' (as in Laravel) since not everyone will have memcached enabled (especially in windows)

### DIFF
--- a/config/cache.php
+++ b/config/cache.php
@@ -13,7 +13,7 @@ return [
     |
     */
 
-    'default' => env('CACHE_DRIVER', 'memcached'),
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Hmmm, not sure. I don't think we should encourage the file driver. It's quite limited, and not suited for high volume apis, which is exactly what lumen is for.

sunspikes:
Thank you GrahamCampbell, was just playing around with lumen on windows its superfast :+1: :) but it threw Exception few times because the default cache driver was set to memcache, same for default session storage too.

Issue #70, opened 2015-05-01
route() not including URL parameters - is it intended?


skovmand:
Hi!

First of all thank you for the great work with Lumen!

Having a route, e.g.:
```
$app->get('signin', ['as' => 'signin', 'uses' => 'App\Http\Controllers\SomeController@signIn']);
```

When using ```route('signin', ['username' => 'John']);```, the string returned in Lumen 5.0.8 is ```http://mysite.com/signin``` without the username URL parameter. In Laravel 5, the returned value for the same route is ```http://mysite.com/signin?username=John```

Is this a bug or intended functionality?

GrahamCampbell:
I think the router is intentionally simpler.

thasmo:
Does the `route` function use the router for building the URLs?

PR #71, opened 2015-05-01
Add query strings to URL generator and fix parameter bug

lukasgeiter:
*Issue #70*

This PR makes generating an URL for a named route behave more like in Laravel. Parameters that don't exist in the route itself will be appended as GET parameters.

While implementing this I also found a bug in generating URLs for regex routes. Before this PR route parameters where replaced if the parameter only began with the passed name. For example:

    Route: /foo/{bar} - Name: foo

    route('foo', ['b' => 'hello']);

    => /foo/hello

This is fixed now and only one call to `preg_replace_callback` is needed.

Sorry I couldn't split the bugfix and the other stuff in two PRs since it all concerns the same method...

### DIFF
--- a/src/Routing/UrlGenerator.php
+++ b/src/Routing/UrlGenerator.php
@@ -108,11 +108,17 @@ class UrlGenerator
 
         $uri = $this->app->namedRoutes[$name];
 
-        foreach ($parameters as $key => $value) {
-            $uri = preg_replace('/\{'.$key.'.*?\}/', $value, $uri);
+        $uri = preg_replace_callback('/\{(.*?)(:.*?)?\}/', function ($m) use (&$parameters) {
+            return isset($parameters[$m[1]]) ? array_pull($parameters, $m[1]) : $m[0];
+        }, $uri);
+
+        $uri = $this->to($uri, []);
+
+        if (! empty($parameters)) {
+            $uri .= '?'.http_build_query($parameters);
         }
 
-        return $this->to($uri, []);
+        return $uri;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Routing/UrlGenerator.php
```diff
@@ -108,11 +108,20 @@ public function route($name, $parameters = array())
 
         $uri = $this->app->namedRoutes[$name];
 
-        foreach ($parameters as $key => $value) {
-            $uri = preg_replace('/\{'.$key.'.*?\}/', $value, $uri);
+        $uri = preg_replace_callback('/\{(.*?)(:.*?)?\}/', function($m) use (&$parameters)
+        {
```
GrahamCampbell: please follow psr-2

src/Routing/UrlGenerator.php
```diff
@@ -108,11 +108,20 @@ public function route($name, $parameters = array())
...
+            return isset($parameters[$m[1]]) ? array_pull($parameters, $m[1]) : $m[0];
+
+        }, $uri);
+
+        $uri = $this->to($uri, []);
+
+        if (! empty($parameters))
+        {
```
GrahamCampbell: same

lukasgeiter:
@GrahamCampbell Sorry didn't realize that Lumen was following PSR-2 (already).

GrahamCampbell:
All Laravel projects are following PSR-2 except from laravel itself atm which moves in 3/4 week's tie.

lukasgeiter:
@taylorotwell Thanks for merging! Also don't forget to close #70 :wink:

GrahamCampbell:
@lukasgeiter Thanks. Next time literally include `Closes #70.` in your pull description. Then GitHub will close this issue when the PR is merged. ;)

lukasgeiter:
@GrahamCampbell Oh nice didn't know that. Will do from now on :+1:

PR #75, opened 2015-05-03
Register Hash facade

ffsantos92:
Uncommenting the `$app->withFacades()` call in the bootstrap/app.php file isn't enough to use `Hash` facade.


GrahamCampbell:
@taylorotwell This must have been missed when you ditched lumens hash stuff in favour of laravel's.

PR #78, opened 2015-05-05
Fixed ordering of Application->addRoute as it wasn’t accounting for group prefixes

shnhrrsn:


taylorotwell:
Can you explain this a little more? :smile:

shnhrrsn:
Yep, sorry about that.. Here's an example route:

```php
$app->group([ 'prefix' => 'example' ], function($app) {
	$app->get('/testing', [ 'as' => 'testing', 'action' => function() {
		return 'I am: ' . route('testing');
	}]);
});
```

This will currently output `I am: http://host/testing`, however the real url is `http://host/example/testing`.

In `Application->addRoute` the named route is first added to `namedRoutes` and _then_ the `groupAttributes` are applied, which is where the route's `$uri` var gets updated with the group's prefix.  

By moving the code block that adds the route's uri to `namedRoutes` after the `groupAttributes` block instead of before, the uri is added with the group prefix.

taylorotwell:
Gotcha, thanks <3

PR #88, opened 2015-05-08
add database port to default configuration

intuxicated:
Hi
I want to use postgresql but with different port. seems like lumen configuration accept port option and just it is not there by default. Also we need to add `DB_PORT` to [.env.example](https://github.com/laravel/lumen/blob/master/.env.example) if you accept this PR.

And I have no idea how sql server works, so I leave it to others.


taylorotwell:
This should be sent to 5.0 branch.

Issue #87, opened 2015-05-08
Utilizing Pipeline Earlier In Lifecycle


jasonlewis:
I need to get some feedback here on whether the current approach is ideal in terms of when the pipeline is used and whether or not it should be used earlier.

Currently when the application is run it immediately calls the `dispatch` method, which in turn will check for the existence of a route. Eventually we'll hit the `handleFoundRoute` method if all went well. Now, here is where we first encounter the pipeline. The request is sent through the pipeline and the global middlewares are applied. Then we hit the `handleArrayBasedFoundRoute` method and the pipeline is used once again for the route middlewares.

Now, I'm thinking your reasoning here was to avoid piping through the middlewares too early. Not sure how much of a performance increase you'd get by doing this. I haven't tested. But, would it not make more sense to pipe the request through first, then `dispatch` it. Seems like things are happening backwards here...

My thoughts are this would clean it up a bit. Currently the `prepareResponse` method seems to be called too many times. It actually looks like it's being called on responses that will already be prepared.

If you're curious I'll gladly do a few tests and what not to see how much of a performance hit it takes by piping first.

(Note: The reason for this issue is mainly because this is one of the bigger difference between the routing setups of L5 and Lumen. I've found that L5 pipes first, meaning I can capture the request before the dispatcher even gets to it. Lumen... The route must *exist* before the middlewares are even looked at.)

jasonlewis:
Closing since #89 was merged. :smile:

PR #89, opened 2015-05-09
Tweak Application Routing

jasonlewis:
This is in response to #87.

The idea is that the routing behave in a similar manner to that of L5. Currently, Lumen will look for a route first before it tries to pipe the request through the middlewares.

Laravel 5 will pipe the request through the middlewares first and then find and dispatch the route.

Obviously they both do routing a little differently, however, I think this difference is a bad one and can lead to some unexpected results. As it stands a global middleware in L5 will always be run even if the route does not exist. That's expected behaviour to me. In Lumen, the global middleware will never run if the route does not exist.

This PR fixes that by piping the request through the middleware first.

### Performance

This was going to be my biggest concern, however, it seems the change doesn't affect it at all. Firstly, the request is only piped through the middleware *IF* their are middleware. So it doesn't new up a pipline for no reason. When testing that seemed to be the biggest drain on the amount of requests I could generate per second.

Here are my benchmark results (`ab -t 10 -c 10 http://lumen.app/`):

**Current 5.0 Branch**

Request No. | Requests Per Second
----------------- | --------------------------------
1                   | 222.44
2                   | 225.09
3                   | 229.46
4                   | 224.21
5                   | 224.78
**Avgerage**         | 225.2

**PR Branch**

Request No. | Requests Per Second
----------------- | --------------------------------
1                   | 218.15
2                   | 230.03
3                   | 229.37
4                   | 228.80
5                   | 225.64 
**Avgerage**         | 226.4

(I know the numbers are low... not a lot of memory here!)

The fact is the benchmarks didn't really change at all with the changes I've made.

Signed-off-by: Jason Lewis <jason.lewis1991@gmail.com>

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -1086,13 +1086,15 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
         }
 
         try {
-            if (isset($this->routes[$method.$pathInfo])) {
-                return $this->handleFoundRoute([true, $this->routes[$method.$pathInfo]['action'], []]);
-            }
+            return $this->sendThroughPipeline($this->middleware, function () use ($method, $pathInfo) {
+                if (isset($this->routes[$method.$pathInfo])) {
+                    return $this->handleFoundRoute([true, $this->routes[$method.$pathInfo]['action'], []]);
+                }
 
-            return $this->handleDispatcherResponse(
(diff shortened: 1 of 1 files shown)


jasonlewis:
Awesome, cheers Taylor!

PR #95, opened 2015-05-12
getLocale function added

aydnbhdr:
Some packages uses this function in Laravel, if you want to include them
into your Lumen app, they will crash. This fix solves that problem.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -1500,6 +1500,14 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
     }
 
     /**
+     * @return mixed
+     */
+    public function getLocale()
+    {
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
> Some packages uses this function in Laravel, if you want to include them
into your Lumen app, they will crash. This fix solves that problem.

This function isn't in the app contract, so it's incorrect for them to rely on this function. That said, we could add it in 5.1 to the contact>

aydnbhdr:
sure

taylorotwell:
Laravel packages are not really intended to be compatible with Lumen so I don't want to set a precedent of people expecting them to be or modifying Lumen to insure 100% compatibility with 3rd party Laravel packages.

PR #98, opened 2015-05-13
Update .gitignore

jrean:
Not needed here :)


taylorotwell:
It is sometimes when testing.

Issue #74, opened 2015-05-02
ResponseFactory::download returns empty content body


winglian:
`return response()->download(...);` returns an empty content body. The exact same line in L5 works correctly.

PR #102, opened 2015-05-14
Resolved a problem where response()->download() was returning 0 bytes.

Raistlfiren:
I ran into the same issue as https://github.com/laravel/lumen-framework/issues/74 where the content body was returning nothing. After further debugging I found out that in the Symfony HttpFoundation was returning a maxlen of 0 everytime. The maxlen variable was never being set since the prepare() method was never called. After evoking the prepare() method in BinaryFileResponse it set the maxlen properly and returned the correct size of the file. I don't know if this is the proper way of doing this, but it works.

Thanks!

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -1331,6 +1331,8 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
             $response = new Response($response);
         }
 
+        $response->prepare(Request::capture());
+
         return $response;
     }
 


Arrilot:
Laravel had this bug fixed long time ago. Check https://github.com/laravel/framework/commit/f3ef1653f6f82e2af9c2682a7584d932b974fadd

Notice that response is prepared even if it was just newed up.

Raistlfiren:
Yeah, but the bug is showing up in the Laravel Lumen framework as well. There are several more instances of prepareResponse() in Laravel Lumen framework than the Laravel framework.

taylorotwell:
If you want to use the Symphony PrepareResponse method you should do so in a middleware. I don't want to enable it by default as it is quite slow.

Raistlfiren:
If that is the case, then the Lumen documentation need to be changed to that. Based upon the Lumen documentation it *simply* states - 
screen shot 2015-05-15 at 9 47 24 am

It includes no information based on doing this in middleware. I feel the download() functionality should be either removed from the documentation or additional information should be provided on using it with middleware. Otherwise, the download() functionality will *always* return an empty file, and you will receive other "issues" about it.

Thanks again for your time and review!

taylorotwell:
Why don't you update the PR to just check if it a BinaryFileResponse and
only call prepare() if that is true.

Issue #79, opened 2015-05-05
artisan up & down


jrean:
@taylorotwell would you consider bringing "back" to lumen the `php artisan up` and `php artisan down` command?

Don't know if I'm alone but I'm missing it.
What is the reason you omitted it?

:-)

GrahamCampbell:
This is missed out to keep Lumen light. It you want more stuff, just use Laravel.

PR #104, opened 2015-05-15
Unused Method

jrean:
I asked if we can get the nice artisan command `up` and `down`.
https://github.com/laravel/lumen-framework/issues/79

Since the answer was no, we don't need the `isDownForMaintenance()` method. I checked and it is not called anywhere.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -186,16 +186,6 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
     }
 
     /**
-     * Determine if the application is currently down for maintenance.
-     *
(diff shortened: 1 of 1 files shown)


jrean:
```
PHP Fatal error:  Class Laravel\Lumen\Application contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (Illuminate\Contracts\Foundation\Application::isDownForMaintenance) in /home/travis/build/laravel/lumen-framework/src/Application.php on line 1642
```
Even if we don't use/need that method we need to implement it.

Issue #101, opened 2015-05-14
route doesn't work for encoded url


wutong:
the route doesn't work with encoded url,
like `%E9%A6%96%E9%A1%B5/%E9%A6%96%E9%A1%B5`
could you fix it ?

lukasgeiter:
Please include at least the route declaration. And what does *"doesn't work"* mean? Do you get a 404?

wutong:
@lukasgeiter 
in mycase
I set a route rule

```
$app->get('首页/首页', function()
{
    return 'hello';
}
```
when I type http://xxx.com/首页/首页
It match to `urlencode('首页') . '/' . urlencode('首页')`
not match to `首页/首页`

wutong:
@lukasgeiter 
that's my problem,
could you help me?

BTW: when I use laravel it's works.

wutong:
@lukasgeiter 
I found a solution at 
<https://github.com/nikic/FastRoute/issues/49>
<https://github.com/symfony/symfony/blob/9f855176454aee2065cbba5923288df9f1c7cf91/src/Symfony/Component/Routing/Matcher/UrlMatcher.php#L96>
I just rawdecode the pathinfo~

PR #114, opened 2015-05-23
fix route match bug

yozman:
route match will missed in some urlencode cases

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -1360,7 +1360,9 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
     {
         $query = isset($_SERVER['QUERY_STRING']) ? $_SERVER['QUERY_STRING'] : '';
 
-        return '/'.trim(str_replace('?'.$query, '', $_SERVER['REQUEST_URI']), '/');
+        $path = '/'.trim(str_replace('?'.$query, '', $_SERVER['REQUEST_URI']), '/');
+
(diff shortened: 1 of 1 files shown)


lukasgeiter:
Please add a unit test that would be failing without this change.

yozman:
@lukasgeiter 
I don't know how do show that case.
in my project.
```php
$app->get('你好/中国', function() use ($app) {
    return 'Hello, China';
});
```
than I type `xxx.com/你好/中国` in browser,
it match to `%E4%BD%A0%E5%A5%BD/%E4%B8%AD%E5%9B%BD`
not `你好/中国`. when I use laravel it's doesn't matter.
so I fixed it.

wutong:
@lukasgeiter 
merge it please ~
it fix my problem, [issue](https://github.com/laravel/lumen-framework/issues/101)

lukasgeiter:
@wutong Sorry I can't merge PRs

GrahamCampbell:
@wutong This is unlikely to be merged unless a test case is provided.

taylorotwell:
I can't tell that Symfony HttpFoundation request duplicates this behavior so holding off for now.

yozman:
@taylorotwell 
It doesn't work until now.
I think function `getPathInfo()` at `Application.php`
should return the `urldecode` string,
because `$_SERVER['REQUEST_URI']` is `urlencode` by default.

PR #120, opened 2015-06-01
Added missing dependency to SuperClosure

radmen:
During `Mail::queue()` I've got an exception:

> Fatal error: Class 'SuperClosure\Serializer' not found in /var/www/api/vendor/illuminate/mail/Mailer.php on line 243

It looks like there's a missing dependency to `jeremeamia/superclosure`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,8 @@
         "symfony/http-kernel": "2.6.*",
         "symfony/http-foundation": "2.6.*",
         "symfony/security-core": "2.6.*",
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Not missing. We're not included it on purpose to keep lumen light. This means out of the box, pushing closures to mail isn't supported.

GrahamCampbell:
You can install this dependency in your own app if you really want it.

radmen:
Ah, sorry I missed that in docs.

PR #122, opened 2015-06-01
Fixed getScheme() method

cdarken:
Added property forceSchema to class, replaced $this->request with $this->app->make('request') in getScheme()


cdarken:
Can somebody please take a look at this and confirm or not that it's an issue?

PR #127, opened 2015-06-03
Added access to route definition from request instance

michalkvasnicak:
This pull request solves problem when you want to access route definition in a middleware but `$request->route()` always returns `null`. Now it returns associative array with route definition (action) under key `1`.

```php
$request->route(); // array(..., 1 => ['middleware' => ....]...)
```

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -1159,6 +1159,12 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
 
         $action = $routeInfo[1];
 
+        // set route info as route resolve, so $request->route() will return it
+        // and you can access route info (name, uses, ...) in middleware for example as $request->route()[1]['uses']
+        $this->make('Illuminate\Http\Request')->setRouteResolver(function() use ($routeInfo) {
+            return $routeInfo;
+        });
+
         // Pipe through route middleware...
         if (isset($action['middleware'])) {
             $middleware = $this->gatherMiddlewareClassNames($action['middleware']);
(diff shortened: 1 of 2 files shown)


jurajhrib:
Why has this been closed?

taylorotwell:
It seems like it's already done in registerRequestBindings.

michalkvasnicak:
Binding doesn't seem to work. I've pushed test case for controller too and if you comment out my modification, tests fails.

taylorotwell:
Are you trying to access the route info a global middleware? It will only
be available in route specific middleware.

michalkvasnicak:
We have middleware registered using `$app->routeMiddleware()`.
Also if we define route as `$app->get('/', function(Request $request) { $request->route(); });` it returns null.

In my pull request is test case for this, I've also used xdebug.

michalkvasnicak:
```php
    public function testAccessRequestRouteInController()
    {
        $app = new Application;

        $app->get('/', function (Request $request) {
            if ($request->route()) {
                return response('has route definition');
            }

            return response('Hello World');
        });

        $response = $app->handle(Request::create('/', 'GET'));
        $this->assertEquals(200, $response->getStatusCode());
        $this->assertEquals('has route definition', $response->getContent());
    }
```

this doesn't pass without this modification too

michalkvasnicak:
I've debugged Application and `registerRequestBindings` is not called at all. Just check test case I have in my commit. But when you add my modification, registerRequestBindings gets called.

PR #128, opened 2015-06-04
Return response from welcome

rkgrep:
VerifyCsrfToken breaks at [line 75](https://github.com/laravel/lumen-framework/blob/master/src/Http/Middleware/VerifyCsrfToken.php#L75) because it gets a string instead of response object

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -1592,7 +1592,7 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
      */
     public function welcome()
     {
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
This surely isn't the correct solution.

GrahamCampbell:
We need to fix this lower down so the middleware are always getting a response.

rkgrep:
Didn't think about that.
Ok. I'll refactor this a bit later.

PR #132, opened 2015-06-05
Enable ENV configuration of all db properties

jakubriedl:
This PR adds possibility to configure DB_SCHEMA, DB_CHARSET and DB_COLLATION with ENV properties.


taylorotwell:
This PR has a lot of other old commits :/

PR #133, opened 2015-06-06
Enable ENV configuration of all db

jakubriedl:
enables configuration of DB_SCHEMA, DB_CHARSET and DB_COLLATION using ENV

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -64,8 +64,8 @@ return [
             'database'  => env('DB_DATABASE', 'forge'),
             'username'  => env('DB_USERNAME', 'forge'),
             'password'  => env('DB_PASSWORD', ''),
-            'charset'   => 'utf8',
-            'collation' => 'utf8_unicode_ci',
+            'charset'   => env('DB_CHARSET', 'utf8'),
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
:-1:

jakubriedl:
why not?
This is very important for us to be able to configure DB_SCHEMA for pgsql. Because every micro-service uses different.

GrahamCampbell:
You can configure it, just not using env.

GrahamCampbell:
I think if you want to use env, then use it yourself. Doesn't belong here as a default for everyone IMO.

jakubriedl:
Configuring database schema is the same as setting database name. It is basic argument to connecting to postgres. And anyone who uses multiple schemas in postgres will need it.

Nielsdv:
Hi GrahamCampbell.

Then please remove the hardcoded "public" in schema.
Not only is the mention of the default schema optional in all schema powered databases (so why define it ?) it is also the wrong place to define it, since Postgresql configuration is in charge of the default through the search_path parameter, which in many cases is overriden.

So if this setting does not belong in the env file, it should not contain the arbitrary value of "public" as well.

PR #135, opened 2015-06-08
chg: addRoute method privacy from protected to public

zot24:
We will be able to add our own HTTP Methods if is needed as below:

```
$app->addRoute('PROPFIND', '/principals', 'DavController@principals')
```

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -907,7 +907,7 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
      * @param  string  $uri
      * @param  mixed  $action
      */
-    protected function addRoute($method, $uri, $action)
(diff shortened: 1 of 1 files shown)


zot24:
There is open discussion about this on [Laracast Forum](https://laracasts.com/discuss/channels/lumen/i-need-to-add-propfind-http-method-to-lumen-route-system) where others options where discussed like: extends [Application.app](https://github.com/laravel/lumen-framework/blob/5.0/src/Application.php) to add our own HTTP Methods, like:

```
public function propfind($uri, $action)
{
    $this->addRoute('PROPFIND', $uri, $action);

    return $this;
}
```

However I think this could be a better/cleaner/simpler way to add our own methods if needed, thoughts?

PR #139, opened 2015-06-10
Access current route in controller/route middleware using request->route() method + test

michalkvasnicak:
Added tests too.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -1159,6 +1159,12 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
 
         $action = $routeInfo[1];
 
+        // set route info as route resolve, so $request->route() will return it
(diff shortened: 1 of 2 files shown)


taylorotwell:
This still isn't right. This would resolve the request on every request which is not what we want. There is another way to fix this.

michalkvasnicak:
@taylorotwell I found a problem, there is a line https://github.com/laravel/lumen-framework/blob/5.0/src/Application.php#L1081

This line is preventing application to call registerRequestBindings. If you comment this line out, there is no need to make request as in this pullrequest and tests still passes.

taylorotwell:
​Commenting that out will break other things.​

michalkvasnicak:
Can I ask you what it will break? Because it will be set to `true` here https://github.com/laravel/lumen-framework/blob/5.0/src/Application.php#L363 on `make()` call.

PR #140, opened 2015-06-10
deferred call to boot() method of service providers + test

### LINKED ISSUE #126: Service container is not deferring call of boot methods on service providers
If I register service provider in `bootstrap/app.php`, boot method is called immediately after registration and not after registration of all service providers.

Is this intended behaviour? Because if it is so, then boot method on service provider doesn't make sense and I don't have mechanism to "decorate" services and I have to register service providers in right order.

Example:

```php

// bootstrap/app.php
$app->register('QueryBusServiceProvider');
$app->register('AppServiceProvider');

// QueryBusServiceProvider.php

class QueryBusServiceProvider extends ServiceProvider
{
   public function register()
   {
      $this->app->singleton('QueryBus');
   }
}

// AppServiceProvider.php
class AppServiceProvider extends ServiceProvider
{
   public function boot()
   {
       $queryBus = $this->app->make('QueryBus');
       $queryBus->addExecutor($this->app->make('SomeService'));
   }
}
```

michalkvasnicak:
Call of a `boot()` method on a service provider is deferred until the `dispatch()` is called. 

Fixes #126

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -375,6 +375,8 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
 
     /**
      * Boots service providers
+     *
+     * @return void
      */
     protected function bootServiceProviders()
     {
@@ -382,7 +384,6 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
             $provider->boot();
         }
     }
-
 
     /**
      * Register container bindings for the application.
--- a/src/Application.php
+++ b/src/Application.php
@@ -137,6 +137,13 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
     protected $dispatcher;
 
     /**
+     * Registered service providers that waits to be booted
+     *
+     * @var ServiceProvider[]
+     */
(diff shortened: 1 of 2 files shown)


taylorotwell:
What if you register a service provider after the application has booted now?

michalkvasnicak:
Yes, you are right. Maybe it could call boot method if application is booted otherwise defer it? 

But there is problem if after boot of application I register service provider `A` that has dependency in `boot()` on some services from service provider `B` that is about to be registered too.

michalkvasnicak:
So it just need same mechanism as in laravel otherwise boot methods are useless.

src/Application.php
```diff
@@ -367,6 +374,17 @@ public function make($abstract, $parameters = [])
...
+     * Boots service providers
+     */
+    protected function bootServiceProviders()
+    {
+        foreach ($this->providersToBoot as $provider) {
+            $provider->boot();
+        }
+    }
+
+
```
GrahamCampbell: cs

PR #142, opened 2015-06-11
Fix a error using the 5.1.*@dev

vieirasantosn:
Fix a error using the 5.1.*@dev version missing methods from interface Illuminate\Contracts\Foundation\Application

    public function getCachedCompilePath()
    public function getCachedServicesPath()


taylorotwell:
Lumen has no such paths. I will update for 5.1 soon this month.

vieirasantosn:
Ok, thanks

PR #147, opened 2015-06-16
Fix for fatal error with renamed DispatchesCommands trait

martinssipenko:
PHP Fatal error:  Trait 'Laravel\Lumen\Routing\DispatchesCommands' not found


GrahamCampbell:
You shouldn't be getting a fatal error. That class still exists in L5.1.

GrahamCampbell:
Oh, but Taylor's deleted it in Lumen. :/

PR #166, opened 2015-06-25
[5.1] Fixes #165

### LINKED ISSUE #165: Making Seeder fails
```
$ php artisan make:seeder LinksTableSeeder
PHP Fatal error:  Call to undefined method Laravel\Lumen\Application::getNamespace() in /{path_to_laravel_project}/vendor/illuminate/console/GeneratorCommand.php on line 151

Fatal error: Call to undefined method Laravel\Lumen\Application::getNamespace() in /{path_to_laravel_project}/vendor/illuminate/console/GeneratorCommand.php on line 151
```

The `getNamespace()` method exists in the full-fledged Laravel framework, but not in Lumen. Bug?

Arrilot:
Closes #165

We can hardcode the namespace because there is no `app:name` command in Lumen, right?

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -1538,6 +1538,16 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
     }
 
     /**
+     * Get the application namespace.
+     *
+     * @return string
+     */
+    public function getNamespace()
+    {
+        return 'App\\';
+    }
+
+    /**
      * Set the cached routes.
      *
      * @param  array  $routes
--- a/src/Application.php
+++ b/src/Application.php
@@ -4,6 +4,7 @@ use Closure;
 use Exception;
 use ErrorException;
 use Monolog\Logger;
+use RuntimeException;
 use FastRoute\Dispatcher;
 use Illuminate\Http\Request;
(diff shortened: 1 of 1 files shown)


lukasgeiter:
Just because there's no `app:name` command doesn't mean you can't change the default app namespace.

GrahamCampbell:
Yeh, this looks wrong to me.

Arrilot:
Okay, makes sense.

Let's copy some Laravel stuff then and call it a day.

src/Application.php
```diff
@@ -1538,6 +1556,32 @@ public function getRoutes()
     }
 
     /**
+     * Get the application namespace.
+     *
+     * @return string
+     *
+     * @throws \RuntimeException
+     */
+    public function getNamespace()
```
GrahamCampbell: Why do we even need this method. :-1: for adding it.

Arrilot: As you can see in the issue some Illuminate components that Lumen uses require it .

GrahamCampbell: I don't think we should support those parts.

GrahamCampbell: Infact, we don't...

PR #167, opened 2015-06-25
[5.1] Fix Cache Store Resolution

### LINKED ISSUE #161: Cannot use cache "Class cache.store does not exist"
There appears to be a bug in the Cache component somewhere that prevents it from working on Lumen with the `file` and `array` drivers. (It may affect the others too, I haven't tested then.)

On a fresh install of Lumen, set `CACHE_DRIVER=file` in `.env` and run this route:

```
use Illuminate\Contracts\Cache\Repository;

$app->get('/', function (Repository $cache) use ($app) {
    return $cache->rememberForever('debug', function () {
        return mt_rand(1, 10);
    });
});
```

It should result in this:

```
ReflectionException in Container.php line 736: Class cache.store does not exist
```

I've built a test case that replicates the bug easily: https://github.com/valorin/lumen-cache-bug

```
git clone https://github.com/valorin/lumen-cache-bug
cd lumen-cache-bug/
composer install
php artisan serve
```

Then visit it in your browser, and it should be showing the error.

Keoghan:
Add 'cache.store' binding in Application:registerCacheBindings() - to resolve ReflectionException.

---
Closes #161.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -445,6 +445,10 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
         $this->singleton('cache', function () {
             return $this->loadComponent('cache', 'Illuminate\Cache\CacheServiceProvider');
         });
+        
+        $this->singleton('cache.store', function () {
+            return $this->loadComponent('cache', 'Illuminate\Cache\CacheServiceProvider', 'cache.store');
+        });
     }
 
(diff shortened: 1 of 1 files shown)


valorin:
Awesome, thanks @Keoghan :-)

PR #170, opened 2015-06-30
Return type in comment

fire015:
Set correct return type in docblocks


GrahamCampbell:
Nice catch. :)

fire015:
It's been annoying me for ages!

PR #177, opened 2015-07-09
fix: take withoutMiddlewear() method call in CrawlerTrait to account …

sahanh:
`Laravel\Lumen\Testing\CrawlerTrait` implements `withoutMiddleware` method. However when testing application doesn't skip middleware as expected.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -1357,7 +1357,10 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
      */
     protected function sendThroughPipeline(array $middleware, Closure $then)
     {
-        if (count($middleware) > 0) {
+        $shouldSkipMiddleware = $this->app->bound('middleware.disable') &&
+                                        $this->app->make('middleware.disable') === true;
+
+        if (count($middleware) > 0 and $shouldSkipMiddleware === false) {
             return (new Pipeline($this))
                 ->send($this->make('request'))
                 ->through($middleware)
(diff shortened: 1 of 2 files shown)


src/Application.php
```diff
@@ -1357,7 +1357,10 @@ protected function gatherMiddlewareClassNames($middleware)
      */
     protected function sendThroughPipeline(array $middleware, Closure $then)
     {
-        if (count($middleware) > 0) {
+        $shouldSkipMiddleware = $this->app->bound('middleware.disable') &&
+                                        $this->app->make('middleware.disable') === true;
```
crynobone: Shouldn't this be `$this` instead of `$this->app`

sahanh: Uhm, `$this->app` reference to `$this` through `bootstrapContainer`.

GrahamCampbell: It should be `$this` only.

sahanh:
Updated reference.

ArjanSchouten:
Think you've to squash the 2 commits. Furthermore I think we should use ```&&``` and ```||``` instead of ```and``` and ```or```. In the whole codebase ```&&``` is used.

GrahamCampbell:
Yeh, `and` is NOT a boolean operator. `&&` is.

sahanh:
Do I need to do the squash?

GrahamCampbell:
Yeh, please. :)

PR #180, opened 2015-07-11
Add the option to set a custom name of the cookie for session in the environment file configuration

bradleySuira:
Add the option to set a custom name of the cookie for session in the environment file configuration, by default accept the name of 'laravel_session'

'cookie' => env('SESSION_DRIVER', 'laravel_session'),

### DIFF
--- a/config/session.php
+++ b/config/session.php
@@ -109,7 +109,7 @@ return [
     |
     */
 
-    'cookie' => 'laravel_session',
+    'cookie' => env('SESSION_DRIVER', 'laravel_session'),
 
     /*
     |--------------------------------------------------------------------------


config/session.php
```diff
@@ -109,7 +109,7 @@
     |
     */
 
-    'cookie' => 'laravel_session',
+    'cookie' => env('SESSION_DRIVER', 'laravel_session'),
```
GrahamCampbell: SESSION_DRIVER is definitely wrong here.

bradleySuira: @GrahamCampbell you are right, SESSION_NAME maybe is better ?

My apologize

Issue #107, opened 2015-05-19
Lumen Config Files


sebastiaanluca:
>Lumen will use your copy of the configuration file if you copy and paste one of the files into a config directory within your project root.

Doesn't seem to be the case. `project/config/app.php` exists, but it's not present when doing `config()` or `config('app.key')`. Only when I register the configuration file with `$app->configure('app');` does it successfully load.

Weird thing is, `database.php` _does_ seem to load.

yadakhov:
http://lumen.laravel.com/docs/configuration

>Unlike Laravel, Lumen only uses a single .env configuration file which can be used to configure the various aspects of the framework. 

So there is no configs.  Everything is used for .env.

GrahamCampbell:
> So there is no configs. Everything is used for .env.

What? That's not true at all.

sebastiaanluca:
@GrahamCampbell So Lumen should use my config/app.php instead of the default, right? Same for config/database.php, …

ctrlaltdylan:
Bumping this thread. Even with `$app->configure('any-config-file-in-config-directory');` the application is not aware of any configuration.

`config()->all();` return null.

sebastiaanluca:
@ctrlaltdylan Where exactly did you place that line in your application's flow? I've put mine in the boot method of a loaded service provider and it seems to work, even with custom configuration files (as long as they are in `root/config`, not `app/config`). At the end of the Lumen app file should work too.

ctrlaltdylan:
I placed the configuration file in the application root:

`/config/app.php`

Contents of `/config/app.php`:
```
<?php

return [
    'test' => 'this config should work',
];

In the `/bootstrap/app.php`:

```
$app->configure('app');
```

To test:
```
config('app.test');
```

But only receiving `null`. Even with config's I know are working, like `/config/database.php` and `/config/cache.php`.

sebastiaanluca:
@ctrlaltdylan

- What version of Lumen are you using? I'm using 5.2.
- Did you put `config('app.test');` in a route/controller to test? Has to be __after__ returning `$app` in your `/bootstrap/app.php` file.
- `$app->configure('app');` should be just before `return $app;`.

ctrlaltdylan:
@sebastiaanluca 

Upgrading to 5.2 did the trick. Thank you.

prawnsalad:
[...]

PR #181, opened 2015-07-11
SESSION_COOKIE_NAME env option to set a custom name of the cookie for session

bradleySuira:
Add the option to set a custom name of the cookie for session in the environment file configuration, by default accept the name of 'laravel_session'

'cookie' => env('SESSION_COOKIE_NAME', 'laravel_session')

* Fix the name of env variable in #180

### DIFF
--- a/config/session.php
+++ b/config/session.php
@@ -109,7 +109,7 @@ return [
     |
     */
 
-    'cookie' => 'laravel_session',
+    'cookie' => env('SESSION_DRIVER', 'laravel_session'),
 
     /*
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
I'm :-1: on this. Don't think it should be in the core by default. Why would you want to change this per environment?

bradleySuira:
You are right, related with this issue: https://github.com/laravel/lumen-framework/issues/107, i found the way to set this. 

* create a dir myproject/config/<config-name>.php 
* if exists Lumen load the settings from this.

Thanks

PR #182, opened 2015-07-11
Add post, put and patch test requests with json payload

mikemclin:
I'm using Lumen for an API that accepts JSON payloads. There doesn't appear to be any helpers in the test suite for working with these types of requests.

### DIFF
--- a/src/Testing/CrawlerTrait.php
+++ b/src/Testing/CrawlerTrait.php
@@ -66,6 +66,24 @@ trait CrawlerTrait
     }
 
     /**
+     * Visit the given URI with a POST request and JSON payload.
+     *
+     * @param  string  $uri
+     * @param  mixed  $content
+     * @param  array  $headers
+     * @return $this
+     */
+    public function postJson($uri, $content, array $headers = [])
+    {
+        $server = $this->transformHeadersToServerVars($headers);
+        $json = json_encode($content)
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't think there is really any difference between just passing an array of `$data`.

mikemclin:
These are very different. 

The (main) difference is that the data is being passed into the 7th argument of the [call()](https://github.com/mikemclin/lumen-framework/blob/cd23fdd3b50bd92c19377c909fa2137767bed0e3/src/Testing/CrawlerTrait.php#L678) function (the request body), not the 3rd argument.

Currently to pass a payload in the request body, you have to call the low-level `call()` method, because there are no higher-level methods that allow you to do so.  The current `post()`, `put()` and `patch()` methods actually pass nothing into the 7th argument of the `call()` function.

Perhaps adding a 4th argument to the current methods would be a simpler solution?

Go from this:
```
public function post($uri, array $data = [], array $headers = [])
```

To this:
```
public function post($uri, array $data = [], array $headers = [], $content = null)
```

mikemclin:
This was resolved with this commit: https://github.com/laravel/lumen-framework/commit/365898c7ddbf48e04f206834c06c85c6bb51a5a5#diff-45f07cb7fb874ee64af04c395a31f182R34 where the `json` method was made available and was delivered in 5.2.

Issue #190, opened 2015-07-18
Issue With getPathInfo


GrahamCampbell:
This method is public and relies on super globals. We should either make it protected, or update it not to do that. NB, this method is internally used by lumen to avoid creating a symfony request object.

Issue #186, opened 2015-07-17
Use of PHP superglobals in Application class


AndrewCarterUK:
Hello,

I'm currently developing a series of packages which can be used to create FastCGI applications in PHP. FastCGI applications can stay alive between requests and gain quite serious performance increases. I've already built a couple of packages that allow you to easily convert a Symfony, Slim or Silex application into a FastCGI one - but I came across issues when I tried to do the same with Lumen.

My FastCGI packages rely on being able to provide an application with a Request object (PSR-7 or HttpFoundation) and then receiving a Response object. The use of PHP superglobals in Lumen's Application class (Application::getMethod() and Application::getPathInfo()) obviously cause problems here.

Shouldn't the HTTP method and path information be retrieved from the Request object rather than the Application class?

As far as I can see, the current approach unnecessarily limits the Application class to being able to process one Request-Response cycle per construction?

Anyway - I'd be interested to hear your thoughts are on this. I'm really keen to build a Lumen adapter to my FastCGI libraries but I can't see a clean way forward without making a messy extension to the Application class.

Regards,

Andrew

GrahamCampbell:
If you provide a request object as the first parameter to the `dispatch` method, then we never use those super globals. :)

AndrewCarterUK:
I was more concerned about Lumen apps using those methods.

I assumed Application::getMethod() was public but I've just realised it's protected.

Should Application::getPathInfo() be protected also to prevent apps from using them to obtain information they should retrieve from the request object?

GrahamCampbell:
> Should Application::getPathInfo() be protected also to prevent apps from using them to obtain information they should retrieve from the request object?

Maybe, yeh. Ping @taylorotwell.

AndrewCarterUK:
Should this be reopened whilst discussion of whether Application::getPathInfo() should be public is still ongoing?

GrahamCampbell:
I'll make a new issue. :)

GrahamCampbell:
https://github.com/laravel/lumen-framework/issues/190

PR #191, opened 2015-07-18
Made Application::getPathInfo() protected

AndrewCarterUK:
Implementation of the change suggested and still under discussion here:

https://github.com/laravel/lumen-framework/issues/190
https://github.com/laravel/lumen-framework/issues/186

As people may have built against this method being public - I'm assuming we will have to wait until the next major version before implementing such a change.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -1426,7 +1426,7 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
      *
      * @return string
      */
-    public function getPathInfo()
+    protected function getPathInfo()
     {
         $query = isset($_SERVER['QUERY_STRING']) ? $_SERVER['QUERY_STRING'] : '';
 


taylorotwell:
Breaking change.

lukasgeiter:
PRs for the next release should go to the `master` branch.

GrahamCampbell:
> Breaking change.

I'd argue it's not since that method doesn't work in all cases anyway.

lukasgeiter:
@GrahamCampbell And what about people overriding the method?

GrahamCampbell:
> And what about people overriding the method?

I doubt anyone actually is.

GrahamCampbell:
I'm not overly fussed where this goes tbh as it doesn't technically fix anything. Just prevents people breaking their own code by mistake.

PR #193, opened 2015-07-19
[5.1] Import Exception

### LINKED ISSUE #192: expectEvents is trying to throw Laravel\Lumen\Testing\Exception which doesn't exist
Hello.

When I try to use the `expectEvents()` method and the event wasn't fired, the method is trying to throw a `Laravel\Lumen\Testing\Exception` exception which doesn't exist. The quick fix is to modify the method and add a `\` so it will look for the exception in the global namespace:

```php
public function expectsEvents($events)
    {
        $events = is_array($events) ? $events : func_get_args();

        $mock = Mockery::spy('Illuminate\Contracts\Events\Dispatcher');

        $mock->shouldReceive('fire')->andReturnUsing(function ($called) use (&$events) {
            foreach ($events as $key => $event) {
                if ((is_string($called) && $called === $event) ||
                    (is_string($called) && is_subclass_of($called, $event)) ||
                    (is_object($called) && $called instanceof $event)) {
                    unset($events[$key]);
                }
            }
        });

        $this->beforeApplicationDestroyed(function () use (&$events) {
            if ($events) {
                throw new \Exception(
                    'The following events were not fired: ['.implode(', ', $events).']'
                );
            }
        });

        $this->app->instance('events', $mock);

        return $this;
    }
```

arrilot:
Closes #192

### DIFF
--- a/src/Testing/ApplicationTrait.php
+++ b/src/Testing/ApplicationTrait.php
@@ -1,6 +1,7 @@
 <?php namespace Laravel\Lumen\Testing;
 
 use Mockery;
+use Exception;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Facade;
 use Illuminate\Contracts\Auth\Authenticatable as UserContract;


GrahamCampbell:
This has already been fixed.

arrilot:
Noone is faster than @GrahamCampbell :(

PR #196, opened 2015-07-22
Update session.php to allow naming of session cookie

scottconrad:
This allows the framework to actually change the session cookie name, as the comments directly above it allude to.

### DIFF
--- a/config/session.php
+++ b/config/session.php
@@ -109,7 +109,7 @@ return [
     |
     */
 
-    'cookie' => 'laravel_session',
+    'cookie' => env('COOKIE_NAME','laravel_session'),
 
     /*
     |--------------------------------------------------------------------------


GrahamCampbell:
Just add this in your own app. This is not something someone would normally want to change per environment, it's something you should set in your settings and commit anyway.

scottconrad:
interesting. I thought the .env file was the only place to alter settings, and without the check to env for this I don't see where else to add it.

GrahamCampbell:
> I don't see where else to add it.

You're meant to copy this file into a config folder in your app.

GrahamCampbell:
The env file is ONLY there for changing this specific to your environment.

scottconrad:
oh gotcha. Thanks.

vluzrmos:
Read the docs: http://lumen.laravel.com/docs/installation#configuration-files

PR #204, opened 2015-07-23
Update some files to PSR2

SelimSalihovic:


taylorotwell:
Those don't really look like PSR-2. My linter doesn't change them?

SelimSalihovic:
Used code sniffer to validate. Should probably file an issue there. Take a look at https://gist.github.com/philsturgeon/6320152 - example 3. My bad for not checking the regular psr2 github doc.

GrahamCampbell:
Code sniffer has lots of false positives. That's why we don't use it.

PR #205, opened 2015-07-25
repair custom attributes not working

vanntian:
```php
$attributes = [
            'name' => '姓名',
            'mobile' => '手机',
            'qq' => 'QQ号码',
            'company_name' => '公司名称',
            'company_tel' => '公司电话'
        ];
$this->validate($this->request, $rules, [], $attributes);
```
not working

### DIFF
--- a/src/Routing/ValidatesRequests.php
+++ b/src/Routing/ValidatesRequests.php
@@ -20,7 +20,8 @@ trait ValidatesRequests
     public function validate(Request $request, array $rules, array $messages = [], array $customAttributes = [])
     {
         $validator = $this->getValidationFactory()->make($request->all(), $rules, $messages, $customAttributes);
-         $validator->setAttributeNames($customAttributes);
+        $validator->setAttributeNames($customAttributes);
+        
         if ($validator->fails()) {
             $this->throwValidationException($request, $validator);
         }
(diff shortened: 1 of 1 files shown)


src/Routing/ValidatesRequests.php
```diff
@@ -20,7 +20,7 @@
     public function validate(Request $request, array $rules, array $messages = [], array $customAttributes = [])
     {
         $validator = $this->getValidationFactory()->make($request->all(), $rules, $messages, $customAttributes);
-
+         $validator->setAttributeNames($customAttributes);
```
GrahamCampbell: please fix the indentation

GrahamCampbell: also, you MUST leave one blank line before an if statement

taylorotwell: Why do we even have to do this? We are passing custom attributes into the `make` method?

vanntian: Because when you use ValidatesRequests trait in the controller,if you want change the value of attributes,like this:
```php
$attributes = [
            'name' => '姓名',
            'mobile' => '手机号码',
            'qq' => 'QQ号码',
            'company_name' => '公司名称',
            'company_tel' => '公司电话'
        ];
$this->validate($this->request, $rules, [], $attributes);
```
it's not working

taylorotwell:
This should be unnecessary according to fact that we are passing customAttributes in constructor.

PR #206, opened 2015-07-27
Adds HHVM runtime error code

xiphiaz:
Ref: https://github.com/facebook/hhvm/blob/HHVM-3.8/hphp/runtime/base/runtime-error.h#L62

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -338,7 +338,7 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
      */
     protected function isFatalError($type)
     {
-        return in_array($type, [E_ERROR, E_CORE_ERROR, E_COMPILE_ERROR, E_PARSE]);
+        return in_array($type, [E_ERROR, E_CORE_ERROR, E_COMPILE_ERROR, E_PARSE, 16777217]); //16777217 is a fatal error thrown by HHVM
     }
 
     /**
--- a/src/Application.php
+++ b/src/Application.php
@@ -338,7 +338,11 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
      */
     protected function isFatalError($type)
     {
-        return in_array($type, [E_ERROR, E_CORE_ERROR, E_COMPILE_ERROR, E_PARSE, 16777217]); //16777217 is a fatal error thrown by HHVM
+        $errorCodes = [E_ERROR, E_CORE_ERROR, E_COMPILE_ERROR, E_PARSE];
+        if (defined('FATAL_ERROR')){
+            $errorCodes[] = FATAL_ERROR; //add the HHVM fatal runtime error to the codes
+        }
+        return in_array($type, $errorCodes);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Can you please use the actual constant.

GrahamCampbell:
Can you do a is defined check, then use it if it is defined.

GrahamCampbell:
We can't just use raw integers like this.

xiphiaz:
@GrahamCampbell yes thats probably a better idea as I first tried using the constant but it wasn't defined in non-hhvm land. Checking if it is defined will work.

src/Application.php
```diff
@@ -338,7 +338,11 @@ protected function registerErrorHandling()
      */
     protected function isFatalError($type)
     {
-        return in_array($type, [E_ERROR, E_CORE_ERROR, E_COMPILE_ERROR, E_PARSE]);
+        $errorCodes = [E_ERROR, E_CORE_ERROR, E_COMPILE_ERROR, E_PARSE];
+        if (defined('FATAL_ERROR')){
```
GrahamCampbell: Please add a blank line before and after this if block.

xiphiaz: @GrahamCampbell ok done.

src/Application.php
```diff
@@ -338,7 +338,13 @@ protected function registerErrorHandling()
      */
     protected function isFatalError($type)
     {
-        return in_array($type, [E_ERROR, E_CORE_ERROR, E_COMPILE_ERROR, E_PARSE]);
+        $errorCodes = [E_ERROR, E_CORE_ERROR, E_COMPILE_ERROR, E_PARSE];
+
+        if (defined('FATAL_ERROR')){
+            $errorCodes[] = FATAL_ERROR; //add the HHVM fatal runtime error to the codes
```
GrahamCampbell: not sure if we need this comment

xiphiaz: @GrahamCampbell ok dropped comment

src/Application.php
```diff
@@ -338,7 +338,13 @@ protected function registerErrorHandling()
      */
     protected function isFatalError($type)
     {
-        return in_array($type, [E_ERROR, E_CORE_ERROR, E_COMPILE_ERROR, E_PARSE]);
+        $errorCodes = [E_ERROR, E_CORE_ERROR, E_COMPILE_ERROR, E_PARSE];
+
+        if (defined('FATAL_ERROR')){
```
taylorotwell: bad CS here.

GrahamCampbell: We could just fix this after merge @taylorotwell?

xiphiaz:
@taylorotwell @GrahamCampbell would it be easier on contributors to enforce your code style preferences with [StyleCI](https://styleci.io/) ?

taylorotwell:
Did Graham make you say that?

xiphiaz:
Hah no I actually use it in my own projects.

xiphiaz:
Oh right I just checked out their repo, didn't realise @GrahamCampbell is the developer of it.

GrahamCampbell:
> would it be easier on contributors to enforce your code style preferences with StyleCI ?

That is the plan, yes. StyleCI isn't quite ready for laravel yet though. It will be in a day or so though. :)

xiphiaz:
:+1:

PR #212, opened 2015-08-03
[5.2] Fixed broken master branch

GrahamCampbell:
This syncs the dependency version constraints with those found in laravel/framework and fixes the exception handler after it was broken by changes made to symfony 2 days ago.


GrahamCampbell:
Looks like the build is failing because we were using a deprecated method that's now deleted.

GrahamCampbell:
See https://github.com/symfony/symfony/pull/15418.

PR #220, opened 2015-08-12
[5.1] Recursively transform data to array

mikemclin:
The `array_sort_recursive()` helper was updated recently (actually the `Str` class method it points to was.  Now it has uncovered this issue (I presume because it is working better than previously).

The `seeJsonContains()` method uses `json_decode` on the response content passing in a value of `true` for the second param, which of course typecasts all objects to arrays.  If the response content is getting all objects typecast to arrays recursively, then the data array passed into the function should get the same treatment.

This wasn't causing an issue before the `array_sort_recursive()` update.  I presume that the nested arrays are now getting sorted properly and causing the objects not to match.

### DIFF
--- a/src/Testing/CrawlerTrait.php
+++ b/src/Testing/CrawlerTrait.php
@@ -367,7 +367,7 @@ trait CrawlerTrait
             json_decode($this->response->getContent(), true)
         ));
 
-        foreach (array_sort_recursive($data) as $key => $value) {
+        foreach (array_sort_recursive(json_decode(json_encode($data), true)) as $key => $value) {
             $expected = $this->formatToExpectedJson($key, $value);
 
             $this->assertTrue(


src/Testing/CrawlerTrait.php
```diff
@@ -338,7 +338,7 @@ protected function seeJsonContains(array $data)
             json_decode($this->response->getContent(), true)
         ));
 
-        foreach (array_sort_recursive($data) as $key => $value) {
+        foreach (array_sort_recursive(json_decode(json_encode($data), true)) as $key => $value) {
```
GrahamCampbell: this looks dodgie to me

mikemclin: Not sure if you mean from a code style standpoint or implementation standpoint?  I'm just mimicking the implementation done earlier on in the method to the response data.

```
$actual = json_encode(array_sort_recursive(
    json_decode($this->response->getContent(), true)
));
```

GrahamCampbell: I mean the json_encode/decode hacks.

mikemclin: I agree. I didn't see a Laravel helper function that performs the same functionality (converting all objects, and nested objects to an array recursively).  I mainly just wanted to point out there is an issue, and this is how it can be fixed.  I will take no issue if you want to close this request and implement a more "eloquent" solution.

Currently, you can't have an object anywhere in the data that you pass into this function, because it's keys will not be sorted.  All objects in the response data are converted to arrays, so their keys are sorted.

GrahamCampbell:
Please could you rebase.

taylorotwell:
Can we get a link to the array_sort_recursive change that caused this?

mikemclin:
This is the commit: https://github.com/laravel/framework/commit/35ebb960aa7f5a222f89b814106f7847cffae599

PR #229, opened 2015-08-29
Added route group inheritance

stevijo:
Basically copied over the groupStack functionality from laravel framework with some adjustments. As there are only prefix, middleware and namespace group Attribute.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -97,7 +97,7 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
      *
      * @var array|null
      */
-    protected $groupAttributes;
+    protected $groupStack = [];
 
     /**
      * All of the global middleware for the application.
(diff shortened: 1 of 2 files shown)


src/Application.php
```diff
@@ -895,13 +895,75 @@ public function withEloquent()
...
+     * Format the prefix for the new group attributes.
+     *
+     * @param  array  $new
+     * @param  array  $old
+     * @return string|null
+     */
+    protected static function formatGroupPrefix($new, $old)
+    {
+        $oldPrefix = isset($old['prefix']) ? $old['prefix'] : null;
+        if (isset($new['prefix'])) {
```
GrahamCampbell: missing newline before the if statement

taylorotwell:
I purposefully left nested groups out to keep Lumen slimmer and simpler.

PR #231, opened 2015-08-31
Fix height

Butochnikov:


GrahamCampbell:
Why the master and not 5.1?

GrahamCampbell:
Why not 5.1?

PR #240, opened 2015-09-04
Use an environment variable for cookie name

jamesensembleads:


GrahamCampbell:
This is not normally changed per environment, but per app.

jamesensembleads:
@GrahamCampbell Ok, where in lumen can you configure the cookie name?

GrahamCampbell:
Publish the config to the app.

jamesensembleads:
Thanks

PR #242, opened 2015-09-06
[5.1] Append middleware of action to group middleware

callam:
The middleware of a route should be appended to the middleware of it's group.


GrahamCampbell:
Please rebase.

PR #249, opened 2015-09-20
Provide secure option to UrlGenerator->route() and adhere to forceSchema if applied

jessedc:
I use the singleton UrlGenerator instance to generate links back to my application using named routes and my situation requires generated urls be https regardless of the incoming request protocol.

UrlGenerator does not currently support this functionality via the `route()` method as it does not provide a secure parameter and additionally, it does not honour the `forceSchema()` functionality.

> A case where you would need to override the incoming request protocol would be in a load balanced environment that terminates https at the load balancer. The consumers of an application would be using https, but internally the web servers would receive http messages from the load balancer.

## Quick example

The only way to attempt to force https currently is via `forceSchema()` which is ignored by `route()`.

My expectation of the following code is to return a url with the `https` scheme:

```php
    app('url')->forceSchema('https');
    $selfLink = app('url')->route('cafe', ['id' => 1]);
```

The reason why I'm not getting what I expect is straight forward:

- `route()` internally calls `to()` [without the `$secure` parameter](https://github.com/laravel/lumen-framework/blob/5.1/src/Routing/UrlGenerator.php#L208).
- `to()` uses `getSchemeForUrl(null)` 
- `getSchemeForUrl()` [defers to the current request](https://github.com/laravel/lumen-framework/blob/5.1/src/Routing/UrlGenerator.php#L242) when no $secure value is present.
- `getSchemeForUrl()` does not use [`getScheme()`](https://github.com/laravel/lumen-framework/blob/5.1/src/Routing/UrlGenerator.php#L165) internally, bypassing the functionality provided by `forceSchema()`.

This pull request aims to solve both issues outlined above while reducing duplication and maintaining the same default behavour. 

Cheers!

### DIFF
--- a/src/Routing/UrlGenerator.php
+++ b/src/Routing/UrlGenerator.php
@@ -187,11 +187,12 @@ class UrlGenerator
      *
      * @param  string  $name
      * @param  mixed   $parameters
+     * @param  bool    $secure
      * @return string
      *
      * @throws \InvalidArgumentException
      */
-    public function route($name, $parameters = [])
+    public function route($name, $parameters = [], $secure = null)
     {
         if (! isset($this->app->namedRoutes[$name])) {
             throw new \InvalidArgumentException("Route [{$name}] not defined.");
@@ -205,7 +206,7 @@ class UrlGenerator
             return isset($parameters[$m[1]]) ? array_pull($parameters, $m[1]) : $m[0];
         }, $uri);
 
-        $uri = $this->to($uri, []);
+        $uri = $this->to($uri, [], $secure);
 
         if (! empty($parameters)) {
             $uri .= '?'.http_build_query($parameters);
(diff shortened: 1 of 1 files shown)


src/Routing/UrlGenerator.php
```diff
@@ -187,11 +187,12 @@ public function forceSchema($schema)
      *
      * @param  string  $name
      * @param  mixed   $parameters
+     * @param  bool    $secure
```
GrahamCampbell: bool|null

taylorotwell:
Regarding your use case, you should use the setTrustedProxies method to configure your load balancer as a trusted proxy and then you can send the X-Forwarded-Proto from the load balancer so that you don't really need this feature.

http://symfony.com/doc/current/components/http_foundation/trusting_proxies.html#configuring-header-names

jessedc:
@taylorotwell thanks for the heads up on trusting proxies. Cheers!

PR #250, opened 2015-09-23
Logs : Be abble to change log file path and name

MarcFRICOU:
Hi !
I'm using Lumen as a micro service with Docker. I need to be able to write logs to `php://stdout` to facilitate log aggregation between services. I suggest to add an env. var. `APP_LOG_PATH` default to `storage/logs/lumen.log` as it is hardcoded now.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -641,7 +641,7 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
     protected function getMonologHandler()
     {
         $logPath = env('APP_LOG_PATH', storage_path('logs/lumen.log'));
-        
+
         return (new StreamHandler($logPath, Logger::DEBUG))
                             ->setFormatter(new LineFormatter(null, null, true, true));
     }


src/Application.php
```diff
@@ -640,7 +640,9 @@ protected function registerLogBindings()
      */
     protected function getMonologHandler()
     {
-        return (new StreamHandler(storage_path('logs/lumen.log'), Logger::DEBUG))
+        $logPath = env('APP_LOG_PATH', storage_path('logs/lumen.log'));
```
GrahamCampbell: :-1: for using the env function witihn the core like this.

GrahamCampbell:
This function is protected so you can just extend the application class to override this behaviour.

src/Application.php
MarcFRICOU: I could add a config file. I do understand I can just extend the Application class, but I thought it can be convenient for others.

GrahamCampbell:
Leaving this for Taylor to decide. It's not really up to me.

src/Application.php
MarcFRICOU: the `env` function is already used in the class...

MarcFRICOU:
@taylorotwell Can I know your opinion ? Why does that has to be hard coded ?

jfoliveira:
@taylorotwell @GrahamCampbell 
The documentation at https://lumen.laravel.com/docs/master/errors#logging says:

> By default, Lumen is configured to create daily log files

However, only a single `lumen.log` is being populated.

Looking at `lumen-framework/src/Application.php`:
```php
    protected function getMonologHandler()
    {
        return (new StreamHandler(storage_path('logs/lumen.log'), Logger::DEBUG))
                            ->setFormatter(new LineFormatter(null, null, true, true));
    }
```
It's not reading any configuration file to determine the log handler behavior.
How can I configure Lumen to generate daily log files?

If currently not possible, could you please reopen this PR so we can contribute to make the Lumen logger configurable?

If the documentation is not being honored it would be nice to at least give the PR author an explanation on why his PR is being closed.

maojindao55:
I agree with @MarcFRICOU @jfoliveira ! and why not  make the Lumen logger configurable?

PR #265, opened 2015-10-10
[5.2] Changed $e to $fe in render method.

njohns-grovo:
Good Evening!
I see that tests are currently failing on dev-master.  It appears that either the variable `$e` was meant to be `$fe` or the places in which `$fe` are used should be `$e` instead.

Take care,
Nicholas


GrahamCampbell:
Thanks. My bad there. :)

PR #269, opened 2015-10-16
[5.1] Allow getting path to configuration directory

phanan:
Right now in order to get the configuration directory's path, you need to extend `Application` and access the `protected $appConfigDir` directly. This PR extends `getConfigurationPath()` to accept an empty string and return the config path in the case, and allow public access at the same time.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -844,19 +844,30 @@ class Application extends Container implements ApplicationContract, HttpKernelIn
     }
 
     /**
-     * Get the path to the given configuration file.
+     * Get the path to the given configuration file, or to the directory if the argument is 
+     * empty.
      *
      * @param  string  $name
      * @return string
      */
-    protected function getConfigurationPath($name)
+    public function getConfigurationPath($name = null)
     {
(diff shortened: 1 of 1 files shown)


crynobone:
Can't you just use:

```php
$path = app('path.config');
```

GrahamCampbell:
Yeh.

crynobone:
Eh wait @GrahamCampbell 

This is Lumen, and that not available here.

phanan:
Oh I wish I could...

src/Application.php
```diff
@@ -844,19 +844,30 @@ public function configure($name)
...
     /**
-     * Get the path to the given configuration file.
+     * Get the path to the given configuration file, or to the directory if the argument is 
+     * empty.
      *
      * @param  string  $name
      * @return string
      */
-    protected function getConfigurationPath($name)
+    public function getConfigurationPath($name = '')
```
crynobone: `$name = null` feel more inline with what Laravel normally used.

taylorotwell: Agree.

phanan: Done.

PR #274, opened 2015-10-31
[5.1] Sync lumen's csrf timing safe equals with laravel 5.1

GrahamCampbell:


GrahamCampbell:
NB, this will be updated to just `hash_equals` on the master branch.

PR #279, opened 2015-11-11
Provide ability to register resource routes.

wissit:


taylorotwell:
Holding off for now.

wissit:
Just noticed, I called the function "delete" instead of "destroy". But it doesn't matter now anyway.

PR #280, opened 2015-11-13
Throw a NotFoundHttpException if the route closure is not passed.

jspekken:
When both the closure and the 'uses' key are not passed to the route, the $closure variable stays undefined. This fix will throw a NotFoundHttpException when both are not present.


taylorotwell:
I think it's best to just let it error out here so the developer doesn't think their route is defined correctly.

PR #281, opened 2015-11-13
[5.2] Use database as default auth driver

rkgrep:
Eloquent is disabled by default, so database is more suitable here.


crynobone:
Doesn't withEloquent only load the `db` instance.

rkgrep:
Method name confused me. I always thought it enables models.

Issue #255, opened 2015-10-01
Redirector depends on concrete Laravel\Lumen\Application which has no binding.


ifeltsweet:
`Laravel\Lumen\Http\Redirector` and `Laravel\Lumen\Routing\UrlGenerator` depend on `Laravel\Lumen\Application` to be injected in their constructors.

This breaks application if I in turn inject `Redirector` or `UrlGenerator` because Lumen never bound `Laravel\Lumen\Application` to container.

Lumen binds only the following classes to resolve application:

- `Illuminate\Container\Container`
- `Illuminate\Contracts\Container\Container`
- `Illuminate\Contracts\Foundation\Application`

So, to fix, we should either add `Laravel\Lumen\Application` to be another binding for application or make `Redirector` and `UrlGenerator` use one of the existing bindings.

Stayallive:
To fix this I currently have this is my `bootstrap/app.php`.

```
$app = new Laravel\Lumen\Application(
    realpath(__DIR__ . '/../')
);

$app->singleton('Laravel\Lumen\Application', function() use ($app) {
    return $app;
});
```

kevintanjung:
I had the same problem to when I tried to pass the `Laravel\Lumen\Http\Redirector` instance through a class constructor. It created another instance of `Laravel\Lumen\Application`, because the `Laravel\Lumen\Application` instance is bonded to `Illuminate\Contracts\Foundation\Application`. I also used @Stayallive solution to overcome this, because I rather not overriding the framework behavior.

I think either the attempt above should be done or the `Redirector` or `UrlGenerator` should have instead type-hinted to the Contract. I don't know which one is the better approach though. It does make sense to expect the `Laravel\Lumen\Application` for type-hinting, but it shouldn't make any problem either to let the Contract for type-hinting, because those classes will only be used in junction with Lumen right?

GrahamCampbell:
The bug is simply the binding's not registered. We'd accept a PR to change that.

ifeltsweet:
Just wondering why not use `Illuminate\Contracts\Container\Container` everywhere (like most of the code base does) instead of concrete `Laravel\Lumen\Application` sprinkled here and there.

PR #289, opened 2015-12-13
Change type-hinting of Redirector and URL Generator class

kevintanjung:
So, I try to inject the Redirector instance to a class and I ended up with two instances of `Laravel\Lumen\Application`, because the `Laravel\Lumen\Application` that is initialized through the `bootstrap/app.php` is bound to the `Illuminate\Contracts\Foundation\Application` and not to `Laravel\Lumen\Application`. This is also related to issue #255 

Example:
```php
// bootstrap/app.php
  
<?php
$app->get('/auth', [
    'as' => 'admin.auth',
    'uses' => 'AuthController@admin'
]);
  
  
// app/Http/Middleware/OnlyAllowAdminThrough.php
  
<?php

namespace App\Http\Middleware;

use Laravel\Lumen\Http\Redirector;

class OnlyAllowAdminThrough
{
    public function __construct(Redirector $redirect)
    {
        $this->redirect = $redirect;
    }

    public function handle($request, Closure $next = null)
    {
         if ( ! $request->session->has('is_admin'))
         {
             return $this->redirect->route('admin.auth');
         }

         return $next($request);
    }
}
```

On the Code above, we will not have the Named Route `admin.auth` because a new Lumen Application is instantiated when injecting `Redirector` to the class. For now, to overcome this problem, I just bound the `$app` instance to the `Laravel\Lumen\Application` singleton in `bootstrap/app.php`, but I thought I might send a pull request too, because I believe this behavior is not intended.

### DIFF
--- a/src/Routing/UrlGenerator.php
+++ b/src/Routing/UrlGenerator.php
@@ -2,8 +2,10 @@
 
 namespace Laravel\Lumen\Routing;
 
-use Laravel\Lumen\Application;
+use Illuminate\Contracts\Foundation\Application;
 use Illuminate\Contracts\Routing\UrlRoutable;
+use Laravel\Lumen\Application as Lumen;
+use InvalidArgumentException;
 
 class UrlGenerator
 {
@@ -43,6 +45,10 @@ class UrlGenerator
      */
     public function __construct(Application $app)
     {
+        if (! $app instanceof Lumen) {
+            throw new InvalidArgumentException('The [Laravel\\Lumen\\Routing\\UrlGenerator] requires an instance of Lumen to run.');
+        }
+
         $this->app = $app;
     }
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
Send this to master branch.

src/Routing/UrlGenerator.php
```diff
@@ -43,6 +45,10 @@ class UrlGenerator
      */
     public function __construct(Application $app)
     {
+        if (! $app instanceof Lumen) {
```
GrahamCampbell: this is pointless - just typehint lumen if you actually want it

kevintanjung: Okay, I just add it as fallback, but I don't really think it's required.

So I need to fix this and send to master branch?

PR #290, opened 2015-12-14
Change type-hinting of Redirector and URL Generator class (pull #289)

kevintanjung:
Sending pull #289 to master branch instead of 5.1 branch, also remove checking for Lumen instance.

### DIFF
--- a/src/Routing/UrlGenerator.php
+++ b/src/Routing/UrlGenerator.php
@@ -2,7 +2,7 @@
 
 namespace Laravel\Lumen\Routing;
 
-use Laravel\Lumen\Application;
+use Illuminate\Contracts\Foundation\Application;
 use Illuminate\Contracts\Routing\UrlRoutable;
 
 class UrlGenerator
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,7 @@
         "monolog/monolog": "~1.0",
         "mtdowling/cron-expression": "~1.0",
         "nikic/fast-route": "0.4.*",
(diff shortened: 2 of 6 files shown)


src/Routing/UrlGenerator.php
```diff
@@ -2,7 +2,7 @@
 
 namespace Laravel\Lumen\Routing;
 
-use Laravel\Lumen\Application;
+use Illuminate\Contracts\Foundation\Application;
```
crynobone: `Laravel\Lumen\Application` in 5.2 is no longer implementing `Illuminate\Contracts\Foundation\Application`.

kevintanjung: @taylorotwell told me to send the pull to the master branch, I did push it to 5.1 before. So... what should I do now?

crynobone: Typehint directly to `Laravel\Lumen\Application`

GrahamCampbell:
Please just send your code, not the other stuff too.

kevintanjung:
Yeah, but isn't it because the `master` branch is behind 4 commits from `5.1`? Should I do my code changes on the `master` branch  then?

Moreover, as @crynobone has pointed out, the `master` branch version of `Laravel\Lumen\Application` does not implement the `Illuminate\Contract\Foundation\Application` anymore. If I type-hinted, as I suggest to `5.1`, then it will surely fail.

I don't mind doing the changes but I think we should rather wait what @taylorotwell has to say first right, before I made another PR.

PR #309, opened 2016-01-08
Bump requirement to v5.2.7

crynobone:
`App\Exceptions\Handler` uses `Illuminate\Validation\ValidationException` which is only available starting from 5.2.7.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
         "illuminate/queue": "5.2.*",
         "illuminate/support": "5.2.*",
         "illuminate/translation": "5.2.*",
-        "illuminate/validation": "5.2.*",
+        "illuminate/validation": "~5.2.7",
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
There's no error if that class doesn't exist though, eright?

crynobone:
Can you throw an un-existing exception? https://github.com/laravel/lumen-framework/search?utf8=%E2%9C%93&q=ValidationException

GrahamCampbell:
You only said the exception handler used it....=

crynobone:
> You only said the exception handler used it...

If it's being handled, some code would need to first throws it :trollface:

GrahamCampbell:
> If it's being handled, some code would need to first throws it :trollface:

I assumed that code was in the illuminate packages.

Issue #313, opened 2016-01-09
getRoutes removed. Any alternative for 5.2?


jfoliveira:
[This commit](https://github.com/laravel/lumen-framework/commit/82adca8e67269498387fd0090037e6c40821ef99) removed method `getRoutes` from `Application.php`

How can I programmatically retrieve the list of routes after upgrading to Lumen 5.2?

GrahamCampbell:
What use case did you have for this?

mstaack:
have been using it together with: https://github.com/appzcoder/lumen-route-list

GrahamCampbell:
Hmmm. I'm probably vote not to restore this method. Ping @taylorotwell since you deleted it. ;)

jfoliveira:
@mstaack the mentioned package is also dependent on the `getRoutes` method, see [here](https://github.com/appzcoder/lumen-route-list/blob/1.0/src/RoutesCommand.php#L33), and is therefore unlikely to work with Lumen 5.2.

@GrahamCampbell thanks.
There is no need to restore the method as long as property `routes` in [src/Concerns/RoutesRequests](https://github.com/laravel/lumen-framework/blob/5.2/src/Concerns/RoutesRequests.php#L29) is made public.

Once this property is public one could replace `app()->getRoutes()` with direct access to `app()->routes` property.

jfoliveira:
I sent a new [pull request](#316) bringing the method back.
Better bring the method back than turn `app()->routes` public as having the method existing code remains compatible with 5.2.
Please review the PR.

PR #316, opened 2016-01-10
Brings back app()->getRoutes method

jfoliveira:
The `getRoutes` method has been removed as part of a [code cleanup](https://github.com/laravel/lumen-framework/commit/82adca8e67269498387fd0090037e6c40821ef99).

As it was used by [some Lumen packages](https://github.com/appzcoder/lumen-route-list) this PR brings it back.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -675,16 +675,6 @@ class Application extends Container
     }
 
     /**
-      * Get the raw routes for the application.
-      *
-      * @return array
-      */
-     public function getRoutes()
-     {
-         return $this->routes;
-     }
-
-    /**
      * Register the core container aliases.
      *
      * @return void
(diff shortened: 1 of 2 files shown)


src/Application.php
```diff
@@ -675,6 +675,16 @@ public function prepareForConsoleCommand()
...
 
     /**
+      * Get the raw routes for the application.
+      *
+      * @return array
+      */
+     public function getRoutes()
+     {
+         return $this->routes;
+     }
```
crynobone: This should go under the relevant trait that provides routes.

jfoliveira:
@crynobone you're right. Thanks for pointing that out. Done!

jfoliveira:
fix #313

PR #317, opened 2016-01-10
Fix app language path

markusjwetzel:


src/Application.php
```diff
@@ -457,7 +457,7 @@ protected function registerTranslationBindings()
      */
     protected function getLanguagePath()
     {
-        if (is_dir($appPath = $this->path().'/resources/lang')) {
+        if (is_dir($appPath = $this->path().'/../resources/lang')) {
```
crynobone: How is this a valid path?

markusjwetzel:
I have created a better fix for this, so this one can be closed.

PR #324, opened 2016-01-12
Fix missing binding for 'cache.store'

dschniepp:
Not possible to use cache on a fresh Lumen installation. ReflectionException “Class cache.store does not exist“ appears due to missing binding for 'cache.store‘. The following pull request should fix the issue.


irazasyed:
Thank you. Was having this issue since the time i installed Lumen 5.2 :+1:

PR #333, opened 2016-01-19
String class references to use ::class

martindilling:
Updated the class references that were written as string to use the class name resolution constant.
The feature was added in PHP 5.5 and helps IDEs recognising the references, making it possible to
jump to references, search for usages, etc.

I don't know if there is a reason it is not used currently, but I couldn't think of a reason so here is
a pull request :)

ref: http://php.net/manual/en/language.oop5.basic.php#language.oop5.basic.class.class

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -94,7 +94,7 @@ class Application extends Container
         static::setInstance($this);
 
         $this->instance('app', $this);
-        $this->instance('Laravel\Lumen\Application', $this);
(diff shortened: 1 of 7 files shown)


GrahamCampbell:
We don't accept PRs like this, sorry.

martindilling:
What do you mean by "like this"? That's not giving me much to work with ;)

Issue #315, opened 2016-01-10
redirect() doesn't work, Lumen 5.2


fengqi:
```
ReflectionException in Container.php line 736:
Class url does not exist
```

I've try to register container bind
```
$app->singleton('url', function ($app) {
    return new Laravel\Lumen\Routing\UrlGenerator($app);
});
```
then:
```
Class session.store does not exist
```

So i try it again:
```
$app->singleton('url', function ($app) {
    return new Laravel\Lumen\Routing\UrlGenerator($app);
});

$app->singleton('session', function ($app) {
    return $app->loadComponent('session', 'Illuminate\Session\SessionServiceProvider');
});

$app->singleton('session.store', function ($app) {
    return $app->loadComponent('session', 'Illuminate\Session\SessionServiceProvider', 'session.store');
});
```

So, is the solution just don't use redirect() ?

GrahamCampbell:
Lumen doesn't support sessions.

GrahamCampbell:
> Class session.store does not exist

Where is this error coming from anyway?

GrahamCampbell:
> So i try it again:

You definitely shouldn't try to bind the session stuff like that.

fengqi:
Yep. 
Redirector and redirectResponse depend session.store.

>Where is this error coming from anyway?

I use is like redirect()->route('some..')->withErrors('some..');

GrahamCampbell:
> Redirector and redirectResponse depend session.store.

Well, either that needs to be changed to make it optional, or it need to be documented that the redirector is unavailable in lumen.

fengqi:
Yes.

darkwinternight:
Right now, redirects are explicitly mentioned in the documentation, yet they don't work:

https://lumen.laravel.com/docs/5.2/responses#redirects

phuongngoc:
I confirm redirect() does not work.
Just simple route:

$app->get('/', function () use ($app) {
    return redirect('/foo');
});

sstringer:
I can confirm redirects do not work in Lumen 5.2, too.

mohislm:
Confirmed: redirect doesn't work in Lumen 5.2.

lapshin1988:
Confirmed also: redirect doesn't work in Lumen 5.2.

maxheckel:
Confirmed as well, redirect does not work.

toocool:
:+1:

yuloh:
@fengqi this is fixed, can you close this issue?

PR #337, opened 2016-01-21
Redirects

matthew-james:
Fixes issue #315.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -469,6 +469,18 @@ class Application extends Container
      *
      * @return void
      */
+    protected function registerUrlGeneratorBindings()
+    {
+        $this->singleton('url', function () {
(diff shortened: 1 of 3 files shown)


dwightwatson:
Would we be able to get a new tagged release with this fix, please? :smile:

sumeeti:
I can see that this fix is present in my files and I still get the error `Call to undefined method Illuminate\Support\Facades\Redirect::to()`.

- Lumen Version: 5.2.9
- PHP Version:7.0.17

I have the following in my bootstrap/app.php
```if (!class_exists('Redirect')) {
    class_alias('Illuminate\Support\Facades\Redirect', 'Redirect');
}```
This is how I want to use it
```public function getRoot()
    {
        return redirect()->to('http://www.gmail.com');
    }```

PR #347, opened 2016-02-02
Do not recreate dispatcher on every dispatch()

HonzaMac:
When I run repeatedly `$app->run($request);` app recreate dispatcher on each `dispatch()`, which have bad behaviour on performance. For me 130ms(now) => to 22ms(fixed).

### DIFF
--- a/src/Concerns/RoutesRequests.php
+++ b/src/Concerns/RoutesRequests.php
@@ -407,7 +407,7 @@ trait RoutesRequests
      */
     protected function createDispatcher()
     {
-        return $this->dispatcher ?: \FastRoute\simpleDispatcher(function ($r) {
+        return $this->dispatcher = $this->dispatcher ?: \FastRoute\simpleDispatcher(function ($r) {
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
This isn't correct surely. If you changed the routes between runs, which should be allowed, this isn't correct. Your persisting information that might not be guaranteed to be correctly the same.

HonzaMac:
I I'm changing routes, I would use `setDispatcher()`

GrahamCampbell:
But this way createDispatcher isn't actually creating a dispatcher?

HonzaMac:
It creates, but just in case it wasn't created before.

GrahamCampbell:
That's not creating then. That's just "make" rather than "create" in laravel terminology.

GrahamCampbell:
Create strictly means "new something", not get something old out from a cache.

HonzaMac:
For our app this change means 100ms down :)
So, is there another way, how can change this behaviour?

PR #353, opened 2016-02-04
[5.2] Allow route suffix

thybag:
Resubmitted to 5.2 branch as requested in https://github.com/laravel/lumen-framework/pull/352

------

Hello,

We were just looking at creating an API in lumen and noticed it would be really useful if we were able to specify a route suffix (in much the same way as the current route prefixes work). This would allow us to support users adding .xml/.json to the ends of our endpoints without us needing the repeat the regex for every route in our app.

Our own use case for the change looks roughly like:

```php
$app->group([ 'suffix'=> "[.{api_format:json|xml}]" ], function($app){
    $app->get('/foo', 'BlaController@foo');
    $app->get('/bar', 'BlaController@bar');
});
```
Which means `/foo`, `/foo.xml` and `/foo.json` endpoints are all valid for routes within the group.

Thanks,
Carl

### DIFF
--- a/src/Concerns/RoutesRequests.php
+++ b/src/Concerns/RoutesRequests.php
@@ -195,7 +195,7 @@ trait RoutesRequests
             }
 
             if (isset($this->groupAttributes['suffix'])) {
-                $uri  = trim($uri, '/').trim($this->groupAttributes['suffix'], '/');
+                $uri = trim($uri, '/').trim($this->groupAttributes['suffix'], '/');
             }
 
             $action = $this->mergeGroupAttributes($action);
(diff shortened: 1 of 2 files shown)


taylorotwell:
Is it possible to write a test for this?

thybag:
Hello,

Yep - have just added a test for it.

cs462-unikent:
Have just made a small bug fix to allow suffix's beginning with a `/`. 

The suffix now only gets it's slashes on the right trimmed in order to allow suffix's such as `/bla` or `/foo` via  `'suffix'=> "[/{something:foo|bar}]" ]`

I've added an additional test for this use case.

PR #365, opened 2016-02-15
Expand env variables for all configs.

Garbee:
Make as many env variables for config entries as possible. This allows developers to override things without reimplementing the config files in their application for minor edits.

Also some minor code style things for array alignment in the queue config.

### DIFF
--- a/config/broadcasting.php
+++ b/config/broadcasting.php
@@ -37,7 +37,7 @@ return [
 
         'redis' => [
             'driver' => 'redis',
-            'connection' => 'default',
+            'connection' => env('BROADCAST_REDIS_CONNECTION', 'default'),
         ],
 
         'log' => [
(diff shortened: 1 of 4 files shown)


taylorotwell:
There are some inconsistencies here like referring to CACHE_DATABASE_CONNECTION and then DB_COLLSET, etc. (DB VS DATABASE, etc.)

Would like to clean some of that up before merging.

Garbee:
Looks to me like `CACHE_DATABASE_CONNECTION` is the only "inconsistent" one then. I named everything as directly as I could aside from within database where DB is the norm. The scheme I used is, `{file}_{key}_{subkey}` so things are very direct in name compared to the configuration item being edited.

So, renaming that to `CACHE_DB_CONNECTION` seems like it would address the naming issue. Or are there any other items in particular you think are inconsistent?

Garbee:
Also note, `CACHE_DATABASE_TABLE` pre-existed this edit. So then there is internal inconsistency if we change to `CACHE_DB_CONNECTION`.

PR #371, opened 2016-02-22
[5.2] Set Request resolvers when request is provided

matthew-james:
The default route resolver and user resolver were only being set if a request was not passed in.  I.E., the resolver would work if you called `$app->run()`, but not if you called `$app->run($request)`.

This causes `$request->route()` and `$request->user()` to return null when running tests, even though they work fine the rest of the time.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -426,12 +426,25 @@ class Application extends Container
     protected function registerRequestBindings()
     {
         $this->singleton('Illuminate\Http\Request', function () {
-            return Request::capture()->setUserResolver(function () {
-                return $this->make('auth')->user();
-            })->setRouteResolver(function () {
-                return $this->currentRoute;
-            });
+            return $this->configureRequest(Request::capture());
         });
+    }
+
+    /**
+     * Configure the given request instance for use with the application.
+     *
+     * @param \Illuminate\Http\Request $request
+     * @return \Illuminate\Http\Request
+     */
+    protected function configureRequest(Request $request)
+    {
+        $request->setUserResolver(function () {
+            return $this->make('auth')->user();
+        })->setRouteResolver(function () {
+            return $this->currentRoute;
+        });
+
+        return $request;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Concerns/RoutesRequests.php
```diff
@@ -390,12 +390,13 @@ public function dispatch($request = null)
     /**
      * Parse the incoming request and return the method and path info.
      *
+     * @param Request $request
```
GrahamCampbell: please include the FQCN

src/Concerns/RoutesRequests.php
```diff
@@ -390,12 +390,13 @@ public function dispatch($request = null)
...
      * @return array
      */
-    protected function parseIncomingRequest($request)
+    protected function parseIncomingRequest(Request $request = null)
```
GrahamCampbell: adding a typehint here is technically a breaking change

src/Application.php
```diff
@@ -426,15 +426,28 @@ protected function getMonologHandler()
...
-                return $this->currentRoute;
-            });
+            return $this->configureRequest(Request::capture());
         });
     }
 
     /**
+     * Configure the given request instance for use with the application.
+     *
+     * @param Request $request
```
GrahamCampbell: FQCN please

src/Concerns/RoutesRequests.php
matthew-james: Should I remove the typehint? It's not technically necessary for the fix.

GrahamCampbell: Yes please.

GrahamCampbell: The method definition can just be left as is.

GrahamCampbell: Please keep the phpdoc though, but using the FQCN.

matthew-james:
Ok, I updated the phpdoc to use the FQN and removed the unnecessary signature change.  LMK if it needs any more changes.

Thanks Graham!

PR #373, opened 2016-02-25
Allow for giving a connection on which to assert database checks

patrick-radius:
This allows to specify a database connection on which to test database assertions like `seeInDatabase`

### DIFF
--- a/src/Testing/TestCase.php
+++ b/src/Testing/TestCase.php
@@ -120,9 +120,9 @@ abstract class TestCase extends PHPUnit_Framework_TestCase
     /**
      * Assert that a given where condition exists in the database.
      *
-     * @param  string $table
-     * @param  array $data
-     * @param  string $onConnection
+     * @param  string  $table
+     * @param  array  $data
+     * @param  string|null $onConnection
      * @return $this
      */
     protected function seeInDatabase($table, array $data, $onConnection = null)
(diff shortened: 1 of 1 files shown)


src/Testing/TestCase.php
```diff
@@ -120,13 +120,14 @@ public function tearDown()
     /**
      * Assert that a given where condition exists in the database.
      *
-     * @param  string  $table
-     * @param  array  $data
+     * @param  string $table
```
GrahamCampbell: Please don't change the alignment.

src/Testing/TestCase.php
```diff
@@ -120,13 +120,14 @@ public function tearDown()
...
+     * @param  array $data
+     * @param  string $onConnection
```
GrahamCampbell: string|null

src/Testing/TestCase.php
```diff
@@ -138,25 +139,27 @@ protected function seeInDatabase($table, array $data)
     /**
      * Assert that a given where condition does not exist in the database.
      *
-     * @param  string  $table
-     * @param  array  $data
+     * @param  string $table
+     * @param  array $data
```
GrahamCampbell: same

src/Testing/TestCase.php
```diff
@@ -138,25 +139,27 @@ protected function seeInDatabase($table, array $data)
...
+        return $this->notSeeInDatabase($table, $data, $onConnection);
     }
 
     /**
      * Assert that a given where condition does not exist in the database.
      *
-     * @param  string  $table
-     * @param  array  $data
+     * @param  string $table
+     * @param  array $data
```
GrahamCampbell: same

PR #385, opened 2016-03-15
Added support to Illuminate/Validator update

helipillo:
Added support to my update of Illuminate/validation/Validator.php (https://github.com/helipillo/validation). 
Which consists in 2 new data validation types which will check if the data given is a date before/after or equal the referenced one.

Changes were:

    - Added error messages for the new validation types: before_equal & after_equal


GrahamCampbell:
We only are going to include things actually merged into our codebase. Thanks though.

Issue #386, opened 2016-03-18
Route groups nesting not working


huoding:
The official document describe that route groups may nesting, but it is not working... https://lumen.laravel.com/docs/5.2/routing#route-group-namespaces

GrahamCampbell:
Thanks for the report, but this looks like a mistake in our docs. We don't support this in Lumen in the interest of pure speed and simplicity, sorry.

PR #389, opened 2016-03-20
[5.2] Fix not being to able nest group attributes

kevintanjung:
So the documentation points out that Lumen does support nesting group attributes, although in previous version of Lumen, I think it didn't. @taylorotwell talks about wanting to make the routes faster, so there shouldn't be no complicated route groups nesting. But now it changes?

https://lumen.laravel.com/docs/5.2/routing#route-groups

So I try to implement the nesting of group attributes, I think the problem lies on the line 90, where to preserve the `groupAttributes`, store it temporarily on the `$parentGroupAttributes` variable, but it does not take into account the "previous" parent attributes to the new attributes.

I am not sure whether my fix is the best approach though, because there are some code duplication between the newly added methods, to the `mergeGroupAttributes()` method, which actually merge the attributes into the `action` key of the `routes` information. I think it can be refactor more, but I do not want break too much code yet for now.

Issue #386 .

### DIFF
--- a/src/Concerns/RoutesRequests.php
+++ b/src/Concerns/RoutesRequests.php
@@ -87,11 +87,62 @@ trait RoutesRequests
             $attributes['middleware'] = explode('|', $attributes['middleware']);
         }
 
-        $this->groupAttributes = $attributes;
+        $this->groupAttributes = $this->nestAttributes($attributes);
 
         call_user_func($callback, $this);
 
         $this->groupAttributes = $parentGroupAttributes;
+    }
+
+    /**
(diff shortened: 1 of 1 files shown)


crynobone:
https://github.com/laravel/lumen-framework/pull/229#issuecomment-136025105

kevintanjung:
Thats what I said too above? But it went from in lumen 5.1 not documented that is able to nest group to at 5.2 is on documentation that it is able to at least grouped namespace.

taylorotwell:
Holding off on this.

Issue #388, opened 2016-03-20
HTTP Basic Authentication in Lumen 5.2


arubacao:
In Lumen `5.1` I used the out-of-the-box `onceBasic` authentication as described here: [Laravel 5.1 Docs](https://laravel.com/docs/5.1/authentication#stateless-http-basic-authentication)
In Lumen `5.2` I could only find the `onceBasic` method in the `SessionGuard.php`.
Unfortunately i'll get a 
```
Class session.store does not exist
```
trying to authenticate with this guard.
This makes sense since **Sessions aren't available** in Lumen `5.2`.
However, i'd still like to use HTTP Basic Auth for my application.

For now I wrote my own `BasicGuard` which is mainly the `SessionGuard` but without all of the session stuff.
This works great for me, but I'm wondering if there is a **better fix** for this - or maybe this isn't even a problem and i'm missing something?

So my final question is eventually:

**1.  How to authenticate with HTTP Basic Auth _(in Lumen `5.2`)_?
2.  How did the creator(s) intent to transfer the token (for token based auth) from the server to a client _(in Lumen `5.2`)_?**

Cheers,
Chris

GrahamCampbell:
Sorry, we don't process questions on our issue tracker. Feel free to discuss anywhere else though.

arubacao:
Ok.
If someone is interested in my own HTTP Basic Auth Guard:

https://github.com/arubacao/http-basic-auth-guard

PR #390, opened 2016-03-20
Remove 'session' as supported option

arubacao:
Error: `Class session.store does not exist` trying to authenticate with the `session` driver.

Note: `SessionGuard` is the only Guard with a `onceBasic` method

[Related Issue](https://github.com/laravel/lumen-framework/issues/388)


GrahamCampbell:
@taylorotwell Might be good to just remove this option entirely, and have the auth component default to token if there's no config entry?

PR #395, opened 2016-03-27
[5.2] Allow middleware to replace route info and route parameters

sw-double:
This PR resolves route from request's routeResolver after piping it through middleware. It allows for middleware to override array based routeInfo by setting route resolver on request instance.

For example it makes possible to use implicit route model binding middleware like this (code taken from Lumen's big brother and adapted for array based route): [SubstituteImplicitBindings.php](https://gist.github.com/sw-double/bdc8285b61be5a8e064d)

### DIFF
--- a/src/Concerns/RoutesRequests.php
+++ b/src/Concerns/RoutesRequests.php
@@ -470,8 +470,8 @@ trait RoutesRequests
         if (isset($action['middleware'])) {
             $middleware = $this->gatherMiddlewareClassNames($action['middleware']);
 
-            return $this->prepareResponse($this->sendThroughPipeline($middleware, function () use ($routeInfo) {
-                return $this->callActionOnArrayBasedRoute($routeInfo);
+            return $this->prepareResponse($this->sendThroughPipeline($middleware, function () {
+                return $this->callActionOnArrayBasedRoute($this['request']->route());
             }));
         }
 


GrahamCampbell:
:-1: from me

GrahamCampbell:
IMO, we should go as far as making the request object immutable.

sw-double:
Why is that? Isn't it one of the use cases of middleware to be able to modify request or response objects? Note that you can easily do this in Laravel, because request object is passed by reference, as opposed to array route currently passed by value in Lumen.

Also absence of implicit model binding is a pain point when migrating API from Laravel to Lumen. I would like to be able to use middleware to implement implicit binding and I think it should be possible to do so.

GrahamCampbell:
> Why is that? Isn't it one of the use cases of middleware to be able to modify request or response objects? 

Modifying responses, of course, but not requests. I have never seen a use case where that's been needed. It's just a hacky mess imo.

sw-double:
I can give you a couple of use cases for modifying requests in middleware off the top of my head.

1. There is a HttpCache by Symfony, which simulates reverse proxy caching (basically like nginx or varnish caching would work). When using this sort of caching you want to increase hit-rate. Let's say that you use utm_* tags for marketing purposes. But in most cases requests for `GET example.com/page` and `GET example.com/page?utm_example=tag` would produce the same content, but the problem is that is will also create duplicate cache objects. You might want to strip utm_ tags _from request object_ in your middleware before it enters Symfony's caching middleware, so that all requests for /page would be served from cache, regardless of utm tags being present in query string. Valid use case for modifying request before it goes deeper into middleware stack don't you think?

2. Ability to implement model binding in middleware which I have already stated above.

GrahamCampbell:
You can implement that without modifying the request object though.

mmghv:
@sw-double If you want to use model binding in Lumen, Check out this package [Lumen Route Binding](https://github.com/mmghv/lumen-route-binding)

It requires :

> php >= 4.5.0
> Lumen 5.*

It supports Explicit binding :

```php
$binder->bind('user', 'App\User');
```

And Implicit binding :

```php
$binder->implicitBind('App\Models');
```

And soon will support `Composite binding` (bind more than one wildcard together, In situations like `posts\{post}\comments\{comment}` you will be able to bind it to a callable that will resolve the `Post` and the `Comment` related to the post)

Also can work with other classes like `Repositories` :

```php
// Add a suffix to the class name
$binder->implicitBind('App\Repositories', '', 'Repository');
```

Can use a custom method on the repository :

```php
// Add a suffix to the class name
$binder->implicitBind('App\Repositories', '', 'Repository', 'findForRoute);
```

Where in the repository class you can do :

```php
public function findForRoute($val)
{
    return $this->model->where('slug', $val)->firstOrFail();
}
```

PR #403, opened 2016-04-19
addRoute now accepts multiple methods

m1:
## Use case

Instead of having to do:
```
$app->options('/', 'Controller@Action');
$app->get('/', 'Controller@Action');
```

Can just do:
```
$app->addRoute(['GET', 'OPTIONS'], '/', 'Controller@Action');
```

### DIFF
--- a/src/Concerns/RoutesRequests.php
+++ b/src/Concerns/RoutesRequests.php
@@ -181,9 +181,9 @@ trait RoutesRequests
     /**
      * Add a route to the collection.
      *
-     * @param  string  $method
-     * @param  string  $uri
-     * @param  mixed  $action
+     * @param  array|string  $method
+     * @param  string        $uri
+     * @param  mixed         $action
      */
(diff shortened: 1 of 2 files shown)


src/Concerns/RoutesRequests.php
```diff
@@ -181,9 +181,9 @@ public function options($uri, $action)
     /**
      * Add a route to the collection.
      *
-     * @param  string  $method
-     * @param  string  $uri
-     * @param  mixed  $action
+     * @param  array|string  $method
+     * @param  string        $uri
```
GrahamCampbell: not sure what all these changes where

GrahamCampbell: see https://github.com/laravel/lumen-framework/commit/52c1208c0fbe062146d4d4697e2aee9da91c8b59

m1: were*

PR #405, opened 2016-04-25
[5.2] Fix undefined variable error on array route

mikakane:
if no uses entry and clousure supplied, then undefined variable error occured.
(eg. misstype uses => use )
 
in this case it should throw Exception like NotFoundHttpException


taylorotwell:
I don't think we want to throw a 404 in this case.

PR #433, opened 2016-06-17
[5.3] Support not registering facade aliases

### LINKED ISSUE #321: Conflict with Pecl Event; can't use artisan console
I have the [PECL Event](http://php.net/manual/en/class.event.php) extension installed.

Because the [`prepareForConsoleCommand`](https://github.com/laravel/lumen-framework/blob/5.2/src/Application.php#L663) calls `withFacades`, The `Event` facade is aliased as `Event`.

This results in a `Cannot redeclare class Event ` exception.

Not sure if this is something you would want to fix, but I figured I would report it in case anyone else has the same issue.

GrahamCampbell:
Note how the default to register them when calling withFacades is true, but the default for the console booting is now false.

---

Closes #321.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -618,12 +618,25 @@ class Application extends Container
     /**
      * Register the facades for the application.
      *
+     * @param  bool  $aliases
      * @return void
      */
-    public function withFacades()
+    public function withFacades($aliases = true)
     {
         Facade::setFacadeApplication($this);
 
+        if ($aliases) {
+            $this->withAliases();
+        }
+    }
+
+    /**
+     * Register the aliases for the application.
+     *
+     * @return void
+     */
+    public function withAliases()
+    {
         if (! static::$aliasesRegistered) {
             static::$aliasesRegistered = true;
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Note that the failing tests are unrelated. I'll fix them seperately.

GrahamCampbell:
See https://github.com/laravel/lumen-framework/pull/434 regarding the fix for those tests.

GrahamCampbell:
#436 is now the only thing stopping this passing.

GrahamCampbell:
Passing now. :)

taylorotwell:
Wouldn't think break code. Why no aliases on console?

GrahamCampbell:
> Wouldn't think break code.

Yeh, it could do, that's why I've targeted 5.3.

> Why no aliases on console?

By forcing people to register these aliases when they don't want to, you are potentially breaking their app due to conflicts with other existing classes, say from PHP extensions. The example from that issue was https://secure.php.net/manual/en/class.event.php.

PR #451, opened 2016-07-24
[5.3] Fix some typehints

iakio:


GrahamCampbell:
These are not wrong though?

iakio:
Yes. Theses are not wrong. But these seem better than original.
I think these changes may provide some benefits for IDE uses.

PR #453, opened 2016-07-25
[5.2] Update MakesHttpRequests.php to have correct seeHeader test output.

christian452:
The method seeHeader in MakesHttpRequests should display the correct "Expected" and "Actual" test output. Right now, it thinks that what comes back in the request is the expected header.


GrahamCampbell:
Thanks for the PR. 👍

christian452:
No problem. Glad to contribute :)

Issue #207, opened 2015-07-28
Unit testing Jobs using mock object always fails


rmblstrp:
Unit testing in Lumen to determine if a Job was dispatched always fails.  Live testing in Lumen shows the job to work as expected.  The unit test also works as expected when moved into a full instance of Laravel.

It appears that *app('Illuminate\Contracts\Bus\Dispatcher')* will return in instance of *Illuminate\Bus\Dispatcher* even though if you inspect *app()->instances* you can see *'Illuminate\Contracts\Bus\Dispatcher' => Mockery_0_Illuminate_Bus_Dispatcher*.  

Calling *app('Illuminate\Contracts\Bus\Dispatcher')* in a full Laravel instance will return *Mockery_0_Illuminate_Bus_Dispatcher* as expected.

**Error:**
Mockery\Exception\InvalidCountException: Method dispatch(object(Mockery\Matcher\Type)) from Mockery_0_Illuminate_Bus_Dispatcher should be called at least 1 times but called 0 times.

Job Example
```php
namespace App\Jobs;
class ExampleJob extends Job
{
    public function handle()
	{
	}
}
```

Unit Test Example
```php
class ExampleTest extends TestCase
{
    public function testBasicExample()
    {
        $this->expectsJobs(App\Jobs\ExampleJob::class);
		$job = new \App\Jobs\ExampleJob();
		app('Illuminate\Contracts\Bus\Dispatcher')->dispatch($job);
    }
}
```

ariponce:
+1

jessedc:
Lumen application's [overloaded `make()` implementation](https://github.com/laravel/lumen-framework/blob/5.1/src/Application.php#L425) is triggering `registerBusBindings()` and binding over the the previously bound mock form `expectsJobs()` in this case because the service binding has not been run prior to the call to `app()` after the mock is bound.

I'm unsure of the best solution, but removing the available binding for the Dispatcher prevents the behaviour:

```php
    public function testBasicExample()
    {
        unset ($this->app->availableBindings['Illuminate\Contracts\Bus\Dispatcher']);       
        $this->expectsJobs(App\Jobs\ExampleJob::class);
        $job = new \App\Jobs\ExampleJob();
        app('Illuminate\Contracts\Bus\Dispatcher')->dispatch($job);
    }
```
You can also trigger the registration prior to setting up the mock by just calling `app('Illuminate\Contracts\Bus\Dispatcher')` somewhere earlier in the test.

[...]

PR #454, opened 2016-07-26
Update TestCase.php to resolve issue #207

christian452:
De-register the Dispatcher bindings so that the mocked version gets bound correctly.

### DIFF
--- a/src/Testing/TestCase.php
+++ b/src/Testing/TestCase.php
@@ -232,6 +232,8 @@ abstract class TestCase extends PHPUnit_Framework_TestCase
     protected function expectsJobs($jobs)
     {
         $jobs = is_array($jobs) ? $jobs : func_get_args();
+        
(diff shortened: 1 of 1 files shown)


christian452:
I'm not sure this is the best way to resolve this problem. If you guys have any feedback I'd be glad to implement a better solution.

taylorotwell:
Can you explain the problem this is trying to solve?

christian452:
I'm having the following issue:
https://github.com/laravel/lumen-framework/issues/207. At the moment the
expectJobs method in TestCase is not working for me. All I'm trying to is
something simple like:

`
public function testJobsDispatch()
{
$this->expectsJobs('App\Jobs\ExampleJob');

dispatch(new ExampleJob());
}
`

and that test throws:

`Method dispatch(object(Mockery\Matcher\Type)) from
Mockery_0_Illuminate_Bus_Dispatcher should be called
 at least 1 times but called 0 times.`

PR #458, opened 2016-08-05
Revert "[5.2] Update MakesHttpRequests.php to have correct seeHeader test output."

GrahamCampbell:
This breaks the assertResponseStatus method.

### DIFF
--- a/src/Testing/Concerns/MakesHttpRequests.php
+++ b/src/Testing/Concerns/MakesHttpRequests.php
@@ -431,7 +431,7 @@ trait MakesHttpRequests
 
         if (! is_null($value)) {
             $this->assertEquals(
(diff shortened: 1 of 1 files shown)


Garbee:
Could you please elaborate on how it is breaking that method?

GrahamCampbell:
Tests fail if you use the assertRresponseStatus method.

GrahamCampbell:
If you revert this commit, they pass. Try it out. :)

Garbee:
> Tests fail 

So seems like the tests should be fixed rather than a revert. As the new parameter order is what is expected.

GrahamCampbell:
I agree, but not on 5.2 since this has breaking consequences for other user provided code that was relying on that order. See what I mean?

Issue #449, opened 2016-07-22
prepare() not called for all Responses


avictor:
As an example, the simple route:

$app->get( '/version', function() {
    return response()->json( '1.0.0' );
});

The client sends request as HTTP/1.1:

GET /version HTTP/1.1
User-Agent: curl/7.35.0
Host: example.com
Accept: _/_

But response generated by Lumen is always HTTP/1.0:

HTTP/1.0 200 OK
Date: Fri, 22 Jul 2016 15:50:08 GMT
Server: Apache
Cache-Control: must-revalidate, no-cache, no-store, private
Pragma: no-cache
Content-Length: 7
Connection: close
Content-Type: application/json

Comparing Laravel to Lumen it seems there is a missing call to prepare() in prepareResponse() in vendor/laravel/lumen-framework/src/Application.php

orottier:
I noticed the same. The protocol switching kills the `Connection: keep-alive` in my setup

driesvints:
Closing this issue because it's already solved, old or not relevant anymore. Feel free to reply if you're still experiencing this issue.

avictor:
I just spun up an AWS instance and tested the latest version of Lumen.  It's still broken.
Version tested:  Lumen (5.7.1) (Laravel Components 5.7.*)


```
> curl -v http://xyz.elasticbeanstalk.com/
* Connected to xyz.elasticbeanstalk.com (34.246.77.31) port 80 (#0)
> GET / HTTP/1.1
> User-Agent: curl/7.35.0
> Host: xyz.elasticbeanstalk.com
> Accept: */*
> 
* HTTP 1.0, assume close after body
< HTTP/1.0 200 OK
< Date: Tue, 02 Oct 2018 12:53:55 GMT
< Server: Apache
< Cache-Control: no-cache, private
< Content-Length: 40
< Connection: close
< Content-Type: text/html; charset=UTF-8
< 
* Closing connection 0
Lumen (5.7.1) (Laravel Components 5.7.*)
```

Request is sent as HTTP/1.1, but Lumen forces response as HTTP/1.0.

This absolutely kills performance since HTTP/1.1 features such as Connection keep-alive cannot be used.

I really should not be forced to hack the vendor code, or add protocol-specific middleware.   The HTTP protocol should be handled by the framework.

driesvints:
Seems to be working ok for me:

```
 ✝ ~/Sites/lumen-test  master±  curl -v http://lumen-test.test
[...]

PR #464, opened 2016-08-18
'Prepare' the HTTP response based on the request before it is sent

orottier:
see #449

I'm pretty sure this is not the right place to perform the response preparation, but could not come up with a better place.

### DIFF
--- a/src/Http/ResponseFactory.php
+++ b/src/Http/ResponseFactory.php
@@ -20,7 +20,8 @@ class ResponseFactory
      */
     public function make($content = '', $status = 200, array $headers = [])
     {
-        return new Response($content, $status, $headers);
+        $response = new Response($content, $status, $headers);
+        return $response->prepare(app()->request);
     }
 
     /**
@@ -38,7 +39,8 @@ class ResponseFactory
             $data = $data->toArray();
         }
 
-        return new JsonResponse($data, $status, $headers, $options);
+        $response = new JsonResponse($data, $status, $headers, $options);
+        return $response->prepare(app()->request);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Http/ResponseFactory.php
```diff
@@ -20,7 +20,9 @@ class ResponseFactory
      */
     public function make($content = '', $status = 200, array $headers = [])
     {
-        return new Response($content, $status, $headers);
+        $response = new Response($content, $status, $headers);
+
+        return $response->prepare(app()->request);
```
GrahamCampbell: This is not the correct way to resolve the current request.

GrahamCampbell:
I think not preparing was intentionally for performance.

orottier:
Interesting.. Not preparing the response means you are stuck on HTTP 1.0 if I read it right
https://github.com/symfony/http-foundation/blob/master/Response.php#L305

Surely you don't want to sacrifise HTTP 1.1 features for this performance optimization?

GrahamCampbell:
We could just force a higher version tbh.

orottier:
Well, actually I don't really understand why the framework is juggling around with HTTP versions. I guess it is a construction to set the HTTP status code and reason phrase in a flexible way.

`prepare` does not look like a very intensive operation btw, just playing around with the headers bag. I could override with a more minimal approach if you wish. So we have some HTTP/1.0 support left

GrahamCampbell:
The intensive operation is building a request object. We don't bother with that normally.

orottier:
I've forced all responses to be sent over HTTP/1.1, which solves the Keep-Alive issues
Please let me know if that fits Lumen requirements!

taylorotwell:
I would suggest using a middleware for this purpose.

orottier:
Hi @taylorotwell, do you mean in the framework or the application code?

PR #465, opened 2016-08-20
route() allow pass parameters as indexed array

mmghv:
allow $parameters to be one value, indexed array or assoc array,
and set them for wildcards by (index or name),

before it had to be passed like :
route('foo', ['wildcard' => 'value']);

but now you can do the following :
route('foo', 'value');
route('foo', ['value']);
route('foo', ['wildcard' => 'value']);
route('foo', ['value1', 'value2']);
route('foo', ['wildcard2' => 'value2', 'wildcard1' => 'value1']);

### DIFF
--- a/src/Routing/UrlGenerator.php
+++ b/src/Routing/UrlGenerator.php
@@ -219,8 +219,14 @@ class UrlGenerator
 
         $parameters = $this->formatParametersForUrl($parameters);
 
+        $index = -1;
+
         $uri = preg_replace_callback('/\{(.*?)(:.*?)?(\{[0-9,]+\})?\}/', function ($m) use (&$parameters) {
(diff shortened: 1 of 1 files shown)


taylorotwell:
Prefer to just pass as an explicit array.

mmghv:
There's nothing wrong with using an explicit array, but it's really annoying when there's only one param which is most cases but still I have to pass an assoc array which the wildcard name.

Also for example I like to use route() to generate most of my URLs in the views so that whenever I change my mind about some routs URLs I only have to change it in the routes file, but with having to explicitly pass the wildcard name in the view although it's only one wildcard, whenever I change this wildcard name, say for example when I made a little extension for Lumen to add route-model-binding feature like Laravel, so I changed most of the resources routes wildcards names, I had to change it in all views too.

PR #479, opened 2016-09-13
Message on NotFoundHttpException and MethodNotAllowedHttpException

astasoft:
## Overview

This PR add message when NotFoundHttpException and MethodNotAllowedHttpException exceptions are thrown.

## Problem

I'm creating an application using Lumen and trying to develop using [Programming by Wishful Thinking](http://sam-serpoosh.github.io/wishful-thinking/) approach which also involving TDD. Problem comes when I started to create non-existent route which part of my wishful and run the unit test via phpunit.

The error message did not tell me which was causing the problem because:
1. Lumen convert all errors into HTTP response even when called from CLI.
2. Even when I throw the error manually to overcome issue number 1 by editing app/Exceptions/Handler.php it still gives me no error message about what was wrong.

Here's my app/Exceptions/Handler.php
```
    /**
     * Render an exception into an HTTP response.
     *
     * @param  \Illuminate\Http\Request  $request
     * @param  \Exception  $e
     * @return \Illuminate\Http\Response
     */
    public function render($request, Exception $e)
    {
        if (env('APP_DEBUG')) {
            $error = sprintf("%s\nFile: %s\n (Line: %s)\nCall Stack:\n%s",
                            $e->getMessage(),
                            $e->getFile(),
                            $e->getLine(),
                            $e->getTraceAsString());
            throw new Exception($error, $e->getCode());
        }

        if ($request->isJson()) {
            return response()->json([
                'code' => $e->getCode(),
                'status' => 'error',
                'message' => $e->getMessage(),
                'data' => $e->getTraceAsString()
            ]);
            return response()->json($response);
        }

        return parent::render($request, $e);
    }
```

## Solution

Throw more meaningful message when non existent route or not allowed method, so it is easier to debug when using TDD on CLI.

### DIFF
--- a/src/Concerns/RoutesRequests.php
+++ b/src/Concerns/RoutesRequests.php
@@ -395,7 +395,9 @@ trait RoutesRequests
                 }
 
                 return $this->handleDispatcherResponse(
-                    $this->createDispatcher()->dispatch($method, $pathInfo)
+                    $this->createDispatcher()->dispatch($method, $pathInfo),
+                    $method,
+                    $pathInfo
                 );
             });
         } catch (Exception $e) {
@@ -452,16 +454,26 @@ trait RoutesRequests
      * Handle the response from the FastRoute dispatcher.
      *
      * @param  array  $routeInfo
+     * @param  string $method
+     * @param  string $pathInfo
      * @return mixed
      */
-    protected function handleDispatcherResponse($routeInfo)
+    protected function handleDispatcherResponse($routeInfo, $method, $pathInfo)
     {
         switch ($routeInfo[0]) {
             case Dispatcher::NOT_FOUND:
-                throw new NotFoundHttpException;
+                throw new NotFoundHttpException(sprintf(
+                    'Route "%s %s" is not found',
+                    $method,
+                    $pathInfo
+                ));
 
             case Dispatcher::METHOD_NOT_ALLOWED:
-                throw new MethodNotAllowedHttpException($routeInfo[1]);
+                throw new MethodNotAllowedHttpException($routeInfo[1], sprintf(
+                    'Method "%s %s" is not allowed',
+                    $method,
+                    $pathInfo
+                ));
 
             case Dispatcher::FOUND:
                 return $this->handleFoundRoute($routeInfo);
--- a/tests/FullApplicationTest.php
+++ b/tests/FullApplicationTest.php
@@ -261,6 +261,30 @@ class FullApplicationTest extends PHPUnit_Framework_TestCase
         $this->assertEquals(404, $response->getStatusCode());
     }
 
+    public function testNotFoundResponseOnDebugMode()
+    {
+        $oldAppDebug = getenv('APP_DEBUG');
+        putenv('APP_DEBUG=true');
+
+        $app = new Application;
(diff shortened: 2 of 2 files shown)


GrahamCampbell:
Exception messages should not reveal internals to users. 👎

astasoft:
@GrahamCampbell, I did not get what you mean. This message only shown when the `APP_DEBUG` is `true`. Without this message when I run phpunit and having undefined route, I do not know where to look for the errors because it return nothing (empty string for the error).

I'm writing a lot of undefined route as part of TDD and fixing them one by one. Without the message I just guessing which part of the code that having undefined route.

What's your suggestion to know what is the error message if I have undefined route? since the Exception object `$e` that already passed to app/Exceptions/Handler.php does not contains any message.

GrahamCampbell:
The convention should be that messages in http exceptions are actually suitable for users to see.

GrahamCampbell:
And all other exceptions are not suitable.

astasoft:
So, in my case what is the proper way to know the errors from console application? since Lumen did not give me details about the error, e.g: undefined route. I need to guess which part of the code that gives me the error.

taylorotwell:
So what error did you receive before this change?

astasoft:
@taylorotwell, Before this change I got this. There is no message in the exception.

```
phpunit tests/acceptance/AddProviderTest.php
PHPUnit 5.5.2 by Sebastian Bergmann and contributors.

E                                                                   1 / 1 (100%)

Time: 101 ms, Memory: 10.00MB

There was 1 error:

1) AddProviderTest::testAddSingleProvider
Exception:
File: /Users/rioastamal/Projects/Sites/projects-which-makes-me-billionaire/api.lintascloud.com/vendor/laravel/lumen-framework/src/Concerns/RoutesRequests.php
 (Line: 466)
Call Stack:
#0 /Users/rioastamal/Projects/Sites/projects-which-makes-me-billionaire/api.lintascloud.com/vendor/laravel/lumen-framework/src/Concerns/RoutesRequests.php(398): Laravel\Lumen\Application->handleDispatcherResponse(Array, 'POST', '/v1/providers')
... [CUT] ...
```

If I turned off the debug mode, which is the case of my API consumer would see, I would got.

```
APP_DEBUG=false phpunit tests/acceptance/AddProviderTest.php
PHPUnit 5.5.2 by Sebastian Bergmann and contributors.

F                                                                   1 / 1 (100%)

Time: 104 ms, Memory: 12.00MB

There was 1 failure:

1) AddProviderTest::testAddSingleProvider
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'Provider ABC successfully added'
+''

/Users/rioastamal/Projects/Sites/projects-which-makes-me-billionaire/api.lintascloud.com/tests/acceptance/AddProviderTest.php:16
```

As you can in both response debug mode true or false - there is no meaningful message that I can use to identify the error.

And this is the code that I use to test -- `test/acceptance/AddProviderTest.php`

```php
<?php

class AddProviderTest extends TestCase
{
    public function testAddSingleProvider()
    {
        $url = '/v1/providers';
        $postData = [
            'user_id' => 1,
            'provider_name' => 'Provider ABC'
        ];
        $this->json('POST', $url, $postData);
        $response = json_decode($this->response->getContent());

        // My wishful code
        $this->assertEquals('Provider ABC successfully added', $response->message);
    }
}
```

taylorotwell:
Instead of checking the actual text in the error message just test that you are receiving the proper HTTP status code.

PR #481, opened 2016-09-14
[5.3] Allows the Error Context to be set within the Handler

GiampaoloFalqui:
Optional context parameter is extremely useful for third party services like Rollbar which allows developers to define the User context of an error, for example.

### DIFF
--- a/src/Exceptions/Handler.php
+++ b/src/Exceptions/Handler.php
@@ -28,10 +28,10 @@ class Handler implements ExceptionHandler
      * Report or log an exception.
      *
      * @param  \Exception  $e
-     * @param  Array       $context
+     * @param  array       $context
      * @return void
      */
-    public function report(Exception $e, Array $context = [])
+    public function report(Exception $e, array $context = [])
     {
         if ($this->shouldntReport($e)) {
             return;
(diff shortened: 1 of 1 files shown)


taylorotwell:
This would be a breaking change. You can still access the current user within the report method using the Auth facade or something.

GiampaoloFalqui:
Even if I can access the report method, I can not use it because the only place where the `send()` function of Monolog is called, is exactly where I've added the context parameter.

Since Lumen handles logging in a different way of Laravel, I cannot set the context independently to the library I am using, but instead I need to push a new handler (of the third-party library) to the `Logger` Monolog instance instantiated in Lumen's IoC.

This is how Rollbar's library registers the Log Listener in Lumen:

```php
$this->app['log']->pushHandler(
    app(RollbarLogHandler::class, [
        $this->app[Rollbar::class]
    ])
);
```

This is how it is registered on Laravel:

```php
$this->app['log']->listen(function ($level, $message, $context) use ($app) {

    if ($user = \Auth::user()) {
        $context['person'] = [
            'id'       => $user->id,
            'username' => $user->username,
            'email'    => $user->email
        ];
    }

    $app[RollbarLogHandler::class]->log($level, $message, $context);
});
```
You can easily see how easier it is passing the context in Laravel.

I don't understand how this should be a breaking change anyway, the `$context` parameter is purposely made optional so that there's nothing to break.

PR #492, opened 2016-09-30
[5.3] Missing trans function

forzadraco:
please help to add missing helper function trans

### DIFF
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -345,3 +345,23 @@ if (! function_exists('view')) {
         return $factory->make($view, $data, $mergeData);
     }
 }
+
+if (! function_exists('trans')) {
+    /**
+     * Translate the given message.
+     *
+     * @param  string  $id
+     * @param  array   $parameters
+     * @param  string  $domain
+     * @param  string  $locale
+     * @return string
+     */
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
I don't think it's "missing". I think it's not included by design?

GrahamCampbell:
Lumen is not designed to render views, where I assume you're using that. It's designed to be used to implement json http apis and the like.

GrahamCampbell:
For real applications, Laravel is recommended, not Lumen.

forzadraco:
Hi Graham,
I understand, but I'm not use view. I'm using lumen for API, there is no view. only json is return.
we need trans function to return response language based on client request. 
for example It's better to not store your localization language file at mobile apps. you need return the error messages in Api level not in Apps level. 

if lumen doesn't have localization feature it is a pity. @taylorotwell what do you think ?

GrahamCampbell:
> if lumen doesn't have localization feature it is a pity

Well, given that this helper works, localization is clearly possible. I was just saying that the helper was probably omitted in order to indicate in an opinionated way to avoid using it.

taylorotwell:
Thanks. For future reference please respect the conventions of the framework. This entire file had methods in alphabetical order, etc.

Issue #239, opened 2015-09-03
Controllers can't be found inside a route group


geocine:
Eg. This will throw an error **Class ServicesController does not exist**

```
$app->group(['prefix' => 'v1' ], function ($app) {
    $app->get('/delete','ServicesController@remove');
});
```

To be able to fix this I must explicitly define the namespace 

```
$app->group(['prefix' => 'v1' , 'namespace' => 'App\Http\Controllers'], function ($app) {
    $app->get('/delete','ServicesController@remove');
});
```

This should not be the case because this is already defined by default on the **bootstrap\app.php**

```
$app->group(['namespace' => 'App\Http\Controllers'], function ($app) {
    require __DIR__.'/../app/Http/routes.php';
});
```

GrahamCampbell:
Yeh, groups don't inherit in lumen, so you need to define the namespace over and over inside nested groups.

jonvargas:
Won't this be fixed?

IanNorris:
I also ran into this. It is not intuitive that this is the case and feels like a bug. At very least the documentation needs to be updated to mention this (if it is already mentioned, it needs mentioning again on the routes page). It is especially confusing as the namespace is set in app.php, so it isn't obvious that routes.php is actually already within a group unless you've dived a bit deeper into how things work.

bogdanghervan:
I also feel that this needs to be better documented.

[This example here](https://lumen.laravel.com/docs/5.2/routing#route-group-namespaces) suggests that namespaces are inherited and a developer should "register controller routes without specifying the full `App\Http\Controllers` namespace prefix.", but this is not the case (at least) with the default namespace group in `bootstrap/app.php`.

I'm defining an additional group in `routes.php` for routes that should get the `auth` middleware. I could apply the middleware to the default namespace group in `bootstrap/app.php`, but that would mean all routes will require authentication and I was hoping to define that more down the road in `routes.php`.

pashamesh:
@bogdanghervan +1

jonwurtzler:
@bogdanghervan +1

najib00:
@bogdanghervan +1
so the documentation not in sync with codes. here i thought that my env is broken.

gokigoks:
What's the resolution for this? its become unintuitive

yuloh:
@geocine looks like this was fixed, can you close this issue?

PR #506, opened 2016-10-17
Makes nested groups inherit previous groups attributes

molteber:
Was working with routes in lumen today and had setup a group with namespace.
Inside that group i was also using another group to set some authentication middleware, and here comes the problem.

When i created routes inside the second group, none of the routes got the attributes from the first group.
The problem is also descripbed in #239 where the second group could not find the (expected) namespace for the controller.

As also mentioned in the same issue, the docs do show that you have to write the full namespace again, but this feels very bloated.

This PR includes
 * Tests for
    * nested `'as' => 'name'`
    * nested `'namespace' => 'Hello'`
    * nested `'middleware' => 'hello|world'`
    * nested `'prefix' => '/hello'`

* Changing how groups works for lumen routes

This will be a breaking change since routes with nested grouping will now behave differently

### DIFF
--- a/src/Concerns/RoutesRequests.php
+++ b/src/Concerns/RoutesRequests.php
@@ -129,7 +129,7 @@ trait RoutesRequests
             $new['as'] = $old['as'].(isset($new['as']) ? '.'.$new['as'] : '');
         }
 
-        if (isset($old['suffix']) && !isset($new['suffix'])) {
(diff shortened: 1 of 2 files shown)


taylorotwell:
Did you copy over the logic from Laravel 5.3?

molteber:
Yeah, with some minor modifications the current group feature for lumen

mstaack:
this is great!

PR #521, opened 2016-11-15
[5.3] Migrations should run first in setUpTraits

gmsantos:
Got this issue when testing on sqlite in memory config for testing.

Looks like Laravel Framework updated in that too https://github.com/laravel/framework/blob/5.3/src/Illuminate/Foundation/Testing/TestCase.php#L99

Maybe this should be target to 5.2 too.


GrahamCampbell:
Didn't we already move this in another PR?

taylorotwell:
Probably.

PR #524, opened 2016-11-25
Adding support for Unix socket of MySQL

myly10:
.env.example PR is also raised


taylorotwell:
Prefer to just keep it matching Laravel.

PR #525, opened 2016-11-29
Quick fix for incorrect routing when app root is in subfolder

phy25:
The current routing handler seems to simplify the getBasePath() function, causing routing problems when the app root is in subfolder.

Introducing optional APP_BASEURL is a quick fix for this, as the full logic of getBasePath in HttpFoundation/Request is really heavy.

Discussion is welcome.


taylorotwell:
The tests are broken. Feel free to submit again when tested.

PR #526, opened 2016-11-29
Quick fix for incorrect routing when app root is in subfolder #525 

phy25:
/* Sorry about the failing tests. Hope that's fixed. */

The current routing handler seems to simplify the getBasePath() function, causing routing problems when the app root is in subfolder.

Introducing optional APP_BASEURL is a quick fix for this, as the full logic of getBasePath in HttpFoundation/Request is really heavy.

Discussion is welcome.

### DIFF
--- a/src/Concerns/RoutesRequests.php
+++ b/src/Concerns/RoutesRequests.php
@@ -692,8 +692,8 @@ trait RoutesRequests
     {
         $query = isset($_SERVER['QUERY_STRING']) ? $_SERVER['QUERY_STRING'] : '';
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Can we break this into multiple lines? It's fairly complicated and hard to read what is happening. Please re-submit once that is done if you want and I will take another look.

phy25:
I will work on this later. However I just don't understand why we need to
open new PRs rather than directly push new commits to this PR 😂

2016年12月8日 上午12:47，"Taylor Otwell" <notifications@github.com>写道：

PR #535, opened 2016-12-16
Render exceptions as JSON if requested

yuloh:
This PR changes the exception handler to render exceptions as JSON if the request specified JSON in the accept header.  I tried to keep the response as similar to the existing HTML response as possible.

If `APP_DEBUG` is false, you just get a code and message.

non-debug response

When `APP_DEBUG` is true you get an extra `data` key with all the exception info.

debug response

If you don't send a JSON accept header you still get the default HTML response.

html response

### DIFF
--- a/src/Exceptions/Handler.php
+++ b/src/Exceptions/Handler.php
@@ -4,6 +4,7 @@ namespace Laravel\Lumen\Exceptions;
 
 use Exception;
 use Illuminate\Http\Response;
+use Illuminate\Http\JsonResponse;
 use Illuminate\Validation\ValidationException;
 use Illuminate\Contracts\Debug\ExceptionHandler;
 use Illuminate\Auth\Access\AuthorizationException;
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
I don't think this should be in the core tbh. Especially since the implementation is quite naive due to the naive implementation of laravel's want's json method. I think we should leave this free for people to implement in their own apps. I have a much more feature rich exceptions package at https://github.com/GrahamCampbell/Laravel-Exceptions, which allows you to perform real content type negociation.

yuloh:
If the implementation of wantsJson is a concern then I would propose that the handler should always return JSON. 

Since Lumen if a framework for building JSON apis I think the current behavior should be changed.  The HTML response is really hard to read and it's a pretty safe assumption the request will be a JSON request.

mstaack:
@yuloh yeah! json in json out!

yuloh:
How about a PR to always return JSON?  Lumen already returns JSON only for validation exceptions.

taylorotwell:
I think I will leave this up to the user for now in their own exception handler. It doesn't feel like this really provides any information you would actually parse and use in your JavaScript.

PR #545, opened 2017-01-06
[5.3] Add support to the secure param on the route helper

### LINKED ISSUE #544: Helper route() don't support the secure param
I'm trying to force the route to https, but the helper doesn't allow that

The route function doesn't pass the "$secure" param to the UrlGenerator instance:

```
function route($name, $parameters = [], $secure = null)
{
    return (new Laravel\Lumen\Routing\UrlGenerator(app()))
                ->route($name, $parameters);
}
```

andergmartins:
Fixes #544

### DIFF
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -287,12 +287,13 @@ if (! function_exists('route')) {
      *
      * @param  string  $name
      * @param  array   $parameters
+     * @param  bool    $secure
      * @return string
      */
-    function route($name, $parameters = [])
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Was going to say please send to 5.4, but of course, it's not possible to override global functions, so there's no bc concerns here. :)

GrahamCampbell:
Looks good to me. 👍

andergmartins:
Nice, thanks

PR #548, opened 2017-01-18
Add doesNotExpectJobs() to TestCase.php

lioannou:
Add `doesNotExpectJobs()`. Basically the inverse of `expectsJobs()`.

Allows you to ensure a specific job(s) is **not** dispatched by your method.

### DIFF
--- a/src/Testing/TestCase.php
+++ b/src/Testing/TestCase.php
@@ -250,6 +250,34 @@ abstract class TestCase extends PHPUnit_Framework_TestCase
     }
 
     /**
+     * Specify a list of jobs that should not be dispatched for the given operation.
+     *
+     * These jobs will be mocked, so that handlers will not actually be executed.
+     *
+     * @param  array|string  $jobs
+     * @return $this
+     */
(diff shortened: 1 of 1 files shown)


taylorotwell:
Usually for these types of things we would say ->withoutJobs, etc. like In Laravel.

lioannou:
But doesn't that just stop jobs from running?

It doesn't test/assert that a job was not actually called?

On 18 Jan 2017, at 19:00, Taylor Otwell <notifications@github.com> wrote:

Usually for these types of things we would say ->withoutJobs, etc. like In
Laravel.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/laravel/lumen-framework/pull/548#issuecomment-273567688>,
or mute the thread

taylorotwell:
Ah OK, you wanted a whole new function. Honestly I would rather just eventually see Laravel's fakes in Lumen. This whole mocking stuff in this situation feels kinda weird and silly (my fault).

PR #558, opened 2017-01-31
Properly create a Request instance to fix tests

themsaid:
While running tests, the `Request` singleton causes the instance registration not to be considered while calling `parseIncomingRequest()`.

In this PR I drop the usage of `singleton()` and register the instance from within `parseIncomingRequest()`, that way the request we get from `Request::capture()` will only be used in normal HTTP requests, while in unit tests the Request we create while using `call, get, post...` will be used.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -422,18 +422,6 @@ class Application extends Container
     }
 
     /**
-     * Register container bindings for the application.
-     *
-     * @return void
-     */
-    protected function registerRequestBindings()
-    {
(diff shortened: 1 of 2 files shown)


oliverlundquist:
Will this be tagged so that I can pull it in with conposer?

GrahamCampbell:
It can still be pulled with composer by using `~5.4.2`. Doesn't have to be tagged.

GrahamCampbell:
Just make sure you have:

```
    "minimum-stability": "dev",
    "prefer-stable": true
```

Issue #565, opened 2017-02-07
lumen 5.4 can't get request headers in serviceProvider


liyu001989:
I want to set my locale,  so I add this code in AppServiceProvider
```
 $request = app('request');
 //dd($request->headers);
 // dd($request->all());
 $acceptLanguage = $request->header('accept-language');
 if ($acceptLanguage) {
     $language = current(explode(',', $acceptLanguage));
     app('translator')->setLocale($language);
 }
```
when I upgrade to 5.4 it's not work, I can't get anything in request, is it a bug?

GrahamCampbell:
Where abouts in your service provider?

liyu001989:
<img width="396" alt="2017-02-07 9 59 24" src="https://cloud.githubusercontent.com/assets/2981799/22694633/efdc7024-ed82-11e6-855d-d9f0d5bee7ac.png">

same code, 5.3 is ok. 5.4 get null

dshepherd:
I’ve just encountered this issue too. It seems that an empty Request object is being returned, with default property values (‘path’ = ‘/‘, method = ‘GET’ etc). 

This didn’t happen in 5.3 and it doesn’t happen in Laravel 5.4.

austenc:
I believe I'm running into the same issue when trying to do a CORS request on Lumen 5.4. It seems the request (using either `app('request')` or the below `$this->app->request`) is a blank object, and not the current request. Same code on lumen 5.3.x returns `OPTIONS` as expected

```
  public function register()
  {
    $request = $this->app->request;

    dd($request->getMethod()); // This is 'GET' when it should be 'OPTIONS' in my case
    ...
  }
```

austenc:
I've done a bit more digging, and it looks like adding this line back (and the `registerRequestBindings` method) fixes the issue I'm having.

https://github.com/laravel/lumen-framework/commit/2ce2eaaaebd315eefcba312de1e568a1b39fcaa5#diff-c9248b3167fc44af085b81db2e292837L874

irazasyed:
Same issue here! I upgraded from 5.2 to 5.4 and it broke this part. I ended up moving the logic to a middleware.

Except10n:
@GrahamCampbell why remove the prepareRequest in initial  Application ?  This leads to serviceprovider can not get a parsed request obj

austenc:
[...]

Issue #546, opened 2017-01-08
Request object dont filled when called in routes.php only during unit tests


magnum-ferox:
Hello! Help me please! Request object dont filled when called in routes.php only during unit tests.
My application listen many hosts and each host can have a custom implementation and unique routes.
I need to know the host name before declare routes. My simplified routes.php:

```php
$Request = app(\Illuminate\Http\Request::class);
if ($Request....) {
// here $Request contains all server variables and headers
// app()->bind(App\Tools\Route\Routes::class, ...) - for specified host implementations
}

// DI for route implementation
$Routes = app(App\Tools\Route\Routes::class);
$Routes->init($app);
```

... and wne called http request (browser) - everything works fine.

I write in unit test for emulate http environment:
`$Response = $this->call('POST', '/', [], [], [], ['someheader' => 'somevalue']);`
But when i run phpunit - $Request in routes.php dont contain server variables and headers!
However i can get my custom headers from Request in controller code.

Because of this: my application cant get $Request variables in routes.php and cant make app()->bind() and cant convigure routes for host. And all test failed.

Please advise how to get headers before routing from tests.

Simple code for reproduce. tests/ExampleTest.php:

```
<?php
use Laravel\Lumen\Testing\DatabaseTransactions;
class ExampleTest extends TestCase {
    public function testExample() {
        $Response = $this->call('POST', '/', [], [], [], ['lalka' => 'kek']);
        var_dump($Response);
    }
}
```

app/Http/routes.php:

```
<?php
$Request = app(\Illuminate\Http\Request::class);
// dd($Request->method()); // contain all variables called http request (browser) and dont contain variables called unit test

$app->post('/', function (\Illuminate\Http\Request $Request) use ($app) {
//        dd($Request->server); // contain all variables called http request (browser) and unit test
    return $app->version();
});
```

GrahamCampbell:
Closing due to inactivity.

PR #568, opened 2017-02-12
Fix Request Binding on initial bootstrap

austenc:
Hello, while trying to work on a new API with lumen 5.4 today I discovered an issue with how the request is bound to the app. I believe this will also take care of #565.

If I have the below code in my lumen app inside a service provider, lumen 5.3 returns 'OPTIONS' (which is correct), whereas 5.4 returns 'GET' -- the default for a blank request. In my case I'm trying to consume my lumen API with vue / axios from another (regular laravel) app.
```
  public function register()
  {
    $request = $this->app->request;

    dd($request->getMethod()); // This is 'GET' when it should be 'OPTIONS' in my case
    ...
  }
```

I noticed that @themsaid made this commit https://github.com/laravel/lumen-framework/commit/2ce2eaaaebd315eefcba312de1e568a1b39fcaa5#diff-c9248b3167fc44af085b81db2e292837L874 recently, so I tried to do some guess and check working backwards from that. Given that I'm not super familiar with the app bootstrap process, I didn't fully understand the intent of @themsaid 's change. That said, there may be a better way to handle this, but my fix seems to work and the tests passed on my local machine!

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -422,6 +422,18 @@ class Application extends Container
     }
 
     /**
+     * Register container bindings for the application.
+     *
+     * @return void
+     */
+    protected function registerRequestBindings()
+    {
+        $this->singleton('Illuminate\Http\Request', function () {
+            return $this->prepareRequest(Request::capture());
+        });
+    }
+
+    /**
      * Prepare the given request instance for use with the application.
      *
      * @param  \Symfony\Component\HttpFoundation\Request $request
@@ -858,6 +870,7 @@ class Application extends Container
         'Illuminate\Contracts\Queue\Queue' => 'registerQueueBindings',
         'Psr\Http\Message\ServerRequestInterface' => 'registerPsrRequestBindings',
         'Psr\Http\Message\ResponseInterface' => 'registerPsrResponseBindings',
+        'Illuminate\Http\Request' => 'registerRequestBindings',
         'translator' => 'registerTranslationBindings',
         'url' => 'registerUrlGeneratorBindings',
         'validator' => 'registerValidatorBindings',


taylorotwell:
@themsaid can you look into this when you have time?

austenc:
This also may relate to #546

themsaid:
@austenc for all request related stuff I recommend you use a global middleware, service providers now run too early while the application haven't captured the request yet.

austenc:
@themsaid This breaks backwards compatibility pretty significantly though. In the case of an `OPTIONS` request, you can't just use a middleware as the bootstrap process will throw a `MethodNotAllowedException` before it hits the middleware. 

Packages like https://github.com/barryvdh/laravel-cors will no longer work as they needed both a service provider and a middleware to accomplish the preflight checks. You can also reference https://gist.github.com/danharper/06d2386f0b826b669552 which does about the same thing as that package. Surely the intention from 5.3 to 5.4 was to maintain backwards compatibility? I'm happy to set up a couple of test apps if you'd like so you can easily reproduce this, just let me know.

themsaid:
@austenc It'll work if you use a global middleware:

```
$app->middleware([
    App\Http\Middleware\CorsMiddleware::class
 ]);
```

In this middleware you'll be able to capture an options request.

austenc:
@themsaid That's what I'm doing, but the middleware doesn't get hit before a `MethodNotAllowed` exception is thrown. I'll make some example apps and link them here

austenc:
Okay, if I basically combine the logic from the service provider and middleware (from the gist i linked), it works. Indeed the middleware is a better way to go. Thanks @themsaid! Is there somewhere we can document this change for Lumen so that others can realize their code needs to be refactored to support the change in 5.4+ ?

themsaid:
If you can open a docs PR on https://github.com/laravel/lumen-docs that would be great :)

austenc:
Will do!

austenc:
Sorry to keep badgering you on this @themsaid but I'm just curious here... shouldn't a change in the app bootstrap process (in this case when the current request is available to the app) be more of a 6.0 change going by semver? Although I agree that it is a good change, it seems like this is something that should've been done in a major release instead.

PR #570, opened 2017-02-12
[5.4] Suggests tinker package

gmsantos:
Like added in https://github.com/laravel/framework/pull/17002/commits/ee8d26cc92db06f01213ac7c1b504610777a4946

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -48,7 +48,8 @@
         "mockery/mockery": "~0.9"
     },
     "suggest": {
-        "vlucas/phpdotenv": "Required to use .env files (~2.2)."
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -48,7 +48,8 @@
         "mockery/mockery": "~0.9"
     },
     "suggest": {
-        "vlucas/phpdotenv": "Required to use .env files (~2.2)."
+        "vlucas/phpdotenv": "Required to use .env files (~2.2).",
+        "laravel/tinker": "Required to use the tinker console command (~1.0).",
```
GrahamCampbell: please remove the extra comma, it's not valid json

PR #575, opened 2017-02-17
Always get the method and path from the Illuminate Request

### LINKED ISSUE #533: Routing doesn't work when deployed to a sub directory
The default [`getPathInfo` method](https://github.com/laravel/lumen-framework/blob/5.3/src/Concerns/RoutesRequests.php#L824) doesn't strip the base url, so deploying to a sub directory like `my-site.com/v1/` will cause the router to attempt to match a request to `/my-site.com/v1/` as `/v1/` instead of `/`.

Reproduction steps:

These steps are a little convoluted since normally public would be set to the web root but they illustrate the problem.

1.) From the root directory of a new lumen app, run `php -S localhost:8000`
2.) Visit `http://localhost:8000/public/` from the browser

Expected output: The default route `/` should be matched.
Actual output: You get a 404 not found.

You can work around the issue by changing the last line of `public/index.php` to `$app->run(\Illuminate\Http\Request::capture());`, since HttpFoundation strips the base url.

yuloh:
Now that #558 is merged, we always create a Request instance before routing.

Since we have a request instance, we can go ahead and use that to get the current method and path, which fixes #533.  Calling `getPathInfo` and `getMethod` only takes ~.1 nanosecond so it shouldn't affect performance.

I also made sure `prepareRequest` is always called so that the route resolver and user resolver are set.

It looks like the path when you weren't using the Request class was trimming the trailing slash, but that doesn't happen when using a Request.  It doesn't look like the full Laravel framework trims trailing slashes so I kept it as is.

### DIFF
--- a/src/Concerns/RoutesRequests.php
+++ b/src/Concerns/RoutesRequests.php
@@ -547,15 +547,13 @@ trait RoutesRequests
      */
     protected function parseIncomingRequest($request)
     {
-        if ($request) {
-            $this->instance(Request::class, $this->prepareRequest($request));
-
-            return [$request->getMethod(), $request->getPathInfo()];
-        } else {
-            $this->instance(Request::class, Request::capture());
-
-            return [$this->getMethod(), $this->getPathInfo()];
+        if (! $request) {
+            $request = Request::capture();
         }
+
+        $this->instance(Request::class, $this->prepareRequest($request));
+
+        return [$request->getMethod(), $request->getPathInfo()];
     }
 
     /**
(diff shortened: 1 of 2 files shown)


taylorotwell:
One of the main points of Lumen is we don't have to create the request instance on every request.

yuloh:
Hey @taylorotwell,

It looks like @themsaid changed Lumen to always create a request in #558.  This PR just uses the existing request for the path and method, and makes sure the request has the user resolver and route resolver set.

PR #580, opened 2017-02-26
fix issue 291

killermonk:
Create a new Request class just for lumen to handle the routing properly.
This Request class extends Laravel's native Request and simply overrides the `route` method.
This allows us to properly handle the array values returned by the lumen's route resolver


tests/FullApplicationTest.php
```diff
@@ -522,6 +522,19 @@ public function testRequestUser()
...
+    {
+        $app = new Application();
+
+        $app->get('/{id}', function (Laravel\Lumen\Http\Request $request) {
+            return $request->route('id');
+        });
+
+        $response = $app->handle(Request::create('/4321', 'GET'));
+
+        $this->assertSame('4321', $response->getContent());
```
killermonk: This is what confuses me:

This test fails with the `Call to a member function setParameter() on array` error before my change.
After my change, if the type hint on the Closure is left as `Illuminate\Http\Request` the error no longer happens, but the test fails because `$response->getContent()` returns ''
With the type hint as `Laravel\Lumen\Http\Request` the test passes.

I've been out of PHP for a few years, so I'm new to the Closure handling in PHP. Not sure what I'm missing with why this won't work with the type hint for the parent class.

taylorotwell:
Not changing this mid-release.

killermonk:
@taylorotwell how would you suggest I make this change? I can put it in master if that is the preferred area. There will be a merge conflict with https://github.com/laravel/lumen-framework/commit/83e65feb5ad436910bfb5cf48e86966612df54ac but it's trivial to resolve. Not sure when the plan is to merge 5.4 into master.

PR #583, opened 2017-03-06
Use PHPUnit Namespace

bubba-h57:
This PR allows upgrading to *PHPUnit 6* for Lumen 5.4 testing, while maintaining backwards compatibility with *PHPUnit 5.7* - it appears that all other references to `PHPUnit_Framework_Assert` have already been migrated and this is the only one left dangling.

In *PHPUnit 6,* PHPUnit's units of code are now namespaced. While *PHPUnit 5.7* has a forward compatibility layer, meaning we can already use `PHPUnit\Framework\{Class}` with that version, there is no backward compatibility layer in *PHPUnit 6* that allows you to use `PHPUnit_Framework_{Class}.`

### DIFF
--- a/src/Testing/Concerns/MakesHttpRequests.php
+++ b/src/Testing/Concerns/MakesHttpRequests.php
@@ -279,7 +279,7 @@ trait MakesHttpRequests
         foreach (array_sort_recursive($data) as $key => $value) {
             $expected = $this->formatToExpectedJson($key, $value);
 
-            call_user_func(['PHPUnit\Framework\Assert', $method],
+            call_user_func([PHPUnit::class, $method],
                 Str::contains($actual, $expected),
                 ($negate ? 'Found unexpected' : 'Unable to find')." JSON fragment [{$expected}] within [{$actual}]."
             );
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
We don't support phpunit 6.x.

GrahamCampbell:
Nore will this change make us actually support it for real.

GrahamCampbell:
Proper support for phpunit 6 will happen in Lumen 5.5.

GrahamCampbell:
> While PHPUnit 5.7 has a forward compatibility layer,

It's forward compatibility layer is only very partial actually. Most things still do not work.

bubba-h57:
You are already doing this everywhere else in the same file. Are you truly arguing against finishing what has been started?

bubba-h57:
I don't think anyone is asking you to *support* PHPunit 6 .... I'm asking for the ability to upgrade to it myself and use it. The `composer.json` remains at the default `5.7`.

GrahamCampbell:
> I don't think anyone is asking you to support PHPunit 6 .... I'm asking for the ability to upgrade to it myself and use it.

Those two are the same things. You can't upgrade to it if we don't support it.

bubba-h57:
I wrongly assumed a mistake had been made when you made the move https://github.com/laravel/lumen-framework/commit/83e65feb5ad436910bfb5cf48e86966612df54ac#diff-45f07cb7fb874ee64af04c395a31f182 from `PHPUnit_Framework_Assert` to `PHPUnit\Framework\Assert`. I just thought you had overlooked it, I didn't realize it was on purpose to prevent the use of PHPUnit 6.

I suppose can just override the `seeJsonContains(array $data, $negate = false)` function in our `tests/TestCase.php` in the meantime. Thanks for the consideration and explanation.

PR #586, opened 2017-03-10
fix issue with sorting of expected value, laravel/internals/issues/248

nattvara:


taylorotwell:
I can't tell what is actually changing?

nattvara:
Try this test

```
public function test() {

	$obj = new \stdClass;
	$obj->foo = '1';
	$obj->bar = '2';
	$obj->baz = '3';

	$arr = [
		'foo' => '1',
		'bar' => '2',
		'baz' => '3'
	];
	
	
	$res = $this->get('/foo')->seeJsonEquals([
            'obj' => $obj,
            'status_code' => 200
        ]);

}
```

```
// route /foo

public function foo(Request $request) {

	$obj = new \stdClass;
	$obj->foo = '1';
	$obj->bar = '2';
	$obj->baz = '3';

	$arr = [
		'foo' => '1',
		'bar' => '2',
		'baz' => '3'
	];

	return response()->json([
		'obj' => $obj,
		'status_code' => 200
	], 200);

}
```


I get this result

```
1) FooTest::test
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'{"obj":{"foo":"1","bar":"2","baz":"3"},"status_code":200}'
+'{"obj":{"bar":"2","baz":"3","foo":"1"},"status_code":200}'
```


Responding with <code>$arr</code> produces the same json in controller and test, but with <code>$obj</code> the order doesn't seem to match

Converting back and forth from JSON in <code>seeJsonEquals</code> with a flag to convert objects to associative arrays seems to fix the sort

I think the problem stems from `array_sort_recursive` using <code>illuminate/support</code> <code>Arr:sortRecursive()</code>

Changing it to this also seems to solve the sorting issue

```
/**
 * Recursively sort an array by keys and values.
 *
 * @param  array  $array
 * @return array
 */
public static function sortRecursive($array)
{
    foreach ($array as &$value) {
        if (is_array($value)) {
            $value = static::sortRecursive($value);
        } elseif (is_object($value)) {
            $value = static::sortRecursive((array) $value); // cast objects to arrays
        }
    }

    if (static::isAssoc($array)) {
        ksort($array);
    } else {
        sort($array);
    }

    return $array;
}
```

PR #588, opened 2017-03-11
[5.4] Use utf8mb4 as mysql default charset

### LINKED ISSUE #587: Use utfmb4 charset as default
laravel uses utfmb4 [by default](https://github.com/laravel/laravel/blob/master/config/database.php#L49-L50)
in laravel document it has been [mentioned](https://github.com/laravel/docs/blob/5.4/migrationas.md#index-lengths--mysql--mariadb)

lumen also should uses this setting as well or at least it should be mentioned in Lumen document.

sijad:
resolves #587


GrahamCampbell:
Probably a better idea to make these changes on 5.5 rather than 5.4, since we don't want to break existing Lumen 5.4 apps.

taylorotwell:
Yeah we can't really make this change now unfortunately. Sorry I missed it. Would have to go on 5.5 now.

PR #606, opened 2017-04-17
[5.4] Validate json fields in convenience method

Cruuncher:
Reasoning for the change:
http://stackoverflow.com/questions/43455562/lumen-validating-json-in-request

### DIFF
--- a/src/Routing/ProvidesConvenienceMethods.php
+++ b/src/Routing/ProvidesConvenienceMethods.php
@@ -58,7 +58,8 @@ trait ProvidesConvenienceMethods
      */
     public function validate(Request $request, array $rules, array $messages = [], array $customAttributes = [])
     {
-        $validator = $this->getValidationFactory()->make($request->all(), $rules, $messages, $customAttributes);
+        $validate_fields = $request->all() + $request->json()->all();
+        $validator = $this->getValidationFactory()->make($validate_fields, $rules, $messages, $customAttributes);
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Breaking change surely?

Cruuncher:
Not sure how. This simply considers JSON in the request as well as form fields. All existing code depending on this will work without trouble as we do an array union(and even prioritize $request->all() so that it doesn't ever get overwritten by json in the request).

laurencei:
Is the existing implementation a bug, on the basis you could unknowingly let unvalidated data through if it came as JSON?

But I agree with @GrahamCampbell - it is a change in behavior that would probably break some peoples apps if they built it on the old behavior (intentionally or not)

laurencei:
I'm confused. I already have Lumen apps that take JSON input via API - and I validate them using `$this->validate($this->request. [])` without any issue?

i.e. I know the JSON fields are being validated - I just triple checked.

So what is this changing?

Cruuncher:
```
    public function test(Request $request) {
        $rules = [
            'test' => 'required',
        ];

        $this->validate($request, $rules);

        return response()->json([
            'status' => 1,
        ]);
    }
```

The following controller is throwing a validation exception given the following json request(and does not post change):
```
{
"test":"test"
}
```

taylorotwell:
Sounds more like you need the proper headers on your incoming request.

PR #612, opened 2017-04-30
Environment variable support for socket file path

Leekas:
Solving the issue I always face in MAMP environment.


taylorotwell:
You an add this to your own application as needed.

PR #621, opened 2017-06-06
Small fix to exception handler

d8vjork:


taylorotwell:
I think the exception handler is OK as-is. Thanks.

PR #623, opened 2017-06-16
Use trusty distribution for testing

vojtasvoboda:
Make Lumen build green again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,14 @@
 language: php
 
-php:
-  - 5.6
-  - 7.0.19
-  - 7.1
-  - hhvm
+matrix:
+  include:
+    - php: 5.6
+    - php: 7.0
+    - php: 7.1
+    - php: hhvm
+      dist: trusty
 
 sudo: false
-
-dist: trusty
 
 install: travis_retry composer install --no-interaction --prefer-source
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: php
 
 php:
   - 5.6
-  - 7.0
+  - 7.0.19
   - 7.1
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Thanks, but this actually results in older versions of php getting pulled in. We only need to activate this on the actual HHVM build.

.travis.yml
```diff
@@ -2,12 +2,14 @@ language: php
 
 php:
   - 5.6
-  - 7.0
+  - 7.0.19
```
GrahamCampbell: Please don't do this. Also, php 5.6 is loaded at an older version too. Just only do dist trusty for hhvm.

GrahamCampbell:
```
matrix:
  include:
    - php: 5.6
    - php: 7.0
    - php: 7.1
    - php: hhvm
      dist: trusty
```

.travis.yml
vojtasvoboda: Yes, that was a bad idea, I just wanted to try if we can force latest PHP version. CircleCI config is updated now.

GrahamCampbell:
> Use CircleCI build matrix

We don't use CircleCI, but meh. :P

vojtasvoboda:
Sorry for typo :-) Now it's ready for merge. Thanks for the help!

taylorotwell:
Why not just make it green without requring trusty?

GrahamCampbell:
> Why not just make it green without requring trusty?

Because Facebook have dropped support for HHVM on that OS. It's not that our tests fail, it's that they don't run.

taylorotwell:
We shouldn't be running tests on HHVM at all I don't think? We don't on the main Laravel repository?

PR #626, opened 2017-06-26
change Application.php __construct function date_default_timezone_set on customize

michealzh:
change reason:
When development or deploy the php application, the devops may be set the php.ini file with config field  `date.timezone` in right value,hens we can check whether the .evn file setted APP_TIMEZONE to change date default_timezone,otherwise We need to change it when __construct at once.


GrahamCampbell:
:-1: I think the current behaviour is more consistent.

PR #631, opened 2017-07-11
Update Handler.php

meanevo:
incorrect style for updated exception page (padding 10px etc...)


meanevo:
I just removed some outdated css, the $css variable here is generated from SymfonyExceptionHandler->getStylesheet(Symfony\Component\Debug).

Laravel does not have this issue as it's calling symfony's debug component directly which has been updated by commit:
https://github.com/symfony/symfony/commit/bab4b9c23c1403b213ed7065c6752aad06c64896#diff-facc3bcdbfaa88c930890ad4713b187dL335

Issue #640, opened 2017-08-03
Not support connectionsToTransact in DatabaseTransactions


Except10n:
- Lumen Version: #.#.#
- PHP Version:
- Database Driver & Version:

### Description:

Lumen DatabaseTransactions
```
public function beginDatabaseTransaction()
    {
        $this->app->make('db')->beginTransaction();

        $this->beforeApplicationDestroyed(function () {
            $this->app->make('db')->rollBack();
        });
    }
```

Laravel DatabaseTransactions
```
public function beginDatabaseTransaction()
    {
        $database = $this->app->make('db');

        foreach ($this->connectionsToTransact() as $name) {
            $database->connection($name)->beginTransaction();
        }

        $this->beforeApplicationDestroyed(function () use ($database) {
            foreach ($this->connectionsToTransact() as $name) {
                $database->connection($name)->rollBack();
            }
        });
    }
```

PR #641, opened 2017-08-03
Support connectionsToTransact feature in Lumen

Except10n:
Add support connectionsToTransact attribute

#640


AnitaYip:
With this change I keep getting `SQLSTATE[08006] [7] FATAL:  sorry, too many clients already` error whenever I run the phpunit tests. Once I revert to lumen version 5.4.6 everything works fine. I'd love to use `connectionsToTransact`, but not sure why I keep getting this error only on this version of the lumen code.

PR #658, opened 2017-09-25
Clear errors that have been handled

### LINKED ISSUE #657: Fatal exceptions are handled twice
- Lumen Version: 5.5
- PHP Version: 7.0.22
- Database Driver & Version: n/a

### Description:
When a controller throws a fatal exception it'll be sent to and reported/rendered by the exception handler twice, duplicating the output.

In my example I have a [`CrashController`](https://github.com/tobiassjosten/lumen-exception-bug/blob/master/app/Http/Controllers/CrashController.php) which triggers a `SoapFault`. It's sent to my exception [`Handler`](https://github.com/tobiassjosten/lumen-exception-bug/blob/master/app/Exceptions/Handler.php) which renders an *"i"*.

Expected output: i
Actual output: ii

This also duplicates any logging you might have in the `report()` method of your handler.

### Steps To Reproduce:
See [lumen-exception-bug](https://github.com/tobiassjosten/lumen-exception-bug) for an example.

tobiassjosten:
Fixes #657

### DIFF
--- a/src/Concerns/RegistersExceptionHandlers.php
+++ b/src/Concerns/RegistersExceptionHandlers.php
@@ -102,6 +102,8 @@ trait RegistersExceptionHandlers
 
         $handler->report($e);
 
+        error_clear_last();
+
         return $handler->render($this->make('request'), $e);
     }
 


taylorotwell:
I don't think this would be the proper solution. @themsaid

taylorotwell:
@themsaid are you able to reproduce this?

themsaid:
@taylorotwell No I can't replicate this, doing a `throw new \Exception('asd');` inside my controller action and the whoops page is rendered only once.

@tobiassjosten can you please guide me on how to replicate this?

tobiassjosten:
The problem is that the error remains in error_get_last(), which gets checked in each and every shutdown. This would clear that because the error has quite literally been handled (sent to handler).

I thought this would be a clean way of solving it without having to restructure too much. Maybe a better approach could be reached if we question why the shutdown handler should check for problems that the error and exception handlers wouldn't catch?

@themsaid Sure! Try instead in your controller, this: `new \SoapClient("asdf.wsdl");`

themsaid:
@tobiassjosten shared screenshots on the issue: https://github.com/laravel/lumen-framework/issues/657#issuecomment-331972661

Still can't replicate.

tobiassjosten:
@themsaid Whaat? 😲 I'm on PHP 7.0.22 but I don't know why it'd make any difference.

Sorry to have you chasing this error – it was so apparent for me, even when I reproduced it in a brand new Lumen app. But could you please check what you're getting from `error_get_last()` in `RegistersExceptionHandlers::handleShutdown()` with that faulty SoapClient?

Issue #663, opened 2017-10-04
Custom Request object is overridden by default Illuminate\Http\Request


deividaspetraitis:
- Lumen Version: 5.5.x
- PHP Version: 7.1
- Database Driver & Version: N/A

### Description:

I have extended default Illuminate\Http\Request object and binded my custom Request to IoC container. When using phpunit default Illuminate\Http\Request is used instead of binded custom request object that causes creating empty Request object without any data and tests to fail.

### Steps To Reproduce:

1. Extend default` Illuminate\Http\Request`
2. Bind custom `Request` to IoC container.
3. Run phpunit tests where request payload is used ( POST, PATCH, etc ).

deividaspetraitis:
Since my PR was rejected as "a breaking change" please see gist to solve this issue by extending and overriding some classes rather editing Lumen`s source: https://gist.github.com/deividaspetraitis/e24d5151bce9102df78c3fc2adda01bd

jgrossi:
Hey @deividaspetraitis I'm experiencing the same issue. I'm using Laravel's API Resource (5.5) in Lumen, and all `$request` I get in `toArray($request)` are empty. Then I have to get the request again from the service container.

I'm gonna take a look in your gist solution. Thanks for that.

PR #664, opened 2017-10-04
Custom Request object is overridden by default Illuminate\Http\Request

deividaspetraitis:
According to https://github.com/laravel/lumen-framework/issues/663

In testing environment `Illuminate\Http\Request` is used and overrides custom `Request` object that is binded to IoC container that makes tests to fail.

### DIFF
--- a/src/Concerns/RoutesRequests.php
+++ b/src/Concerns/RoutesRequests.php
@@ -183,7 +183,7 @@ trait RoutesRequests
             $request = Request::capture();
         }
 
-        $this->instance(Request::class, $this->prepareRequest($request));
+        $this->instance(get_class($request), $this->prepareRequest($request));
 
         return [$request->getMethod(), '/'.trim($request->getPathInfo(), '/')];
     }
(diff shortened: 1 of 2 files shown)


taylorotwell:
Removing the Illuminate\Http\Request binding is a breaking change.

deividaspetraitis:
@taylorotwell  to be more specific binding of `Illuminate\Http\Request` was not removed, please see commit. Fix was about to use binded `Request` object from IoC ( whenever it is `Illuminate\Http\Request` or any other )  indead of *hardcoded* `Illuminate\Http\Request`.

jgrossi:
@taylorotwell I think what @deividaspetraitis is trying to do makes sense. I'm using API Resource in Lumen, and almost all methods inside a `Resource` class, like `with()`, `toArray()` receive a `$request` object, but it's always empty, no route, no path, nothing. This happens only when testing.

I always have to get the request from the service container again:

```php
public function with($request)
{
    $request = app()->make('request');

    return $request->is('auth/login') && $this->jwtToken ?
        ['token' => $this->jwtToken] : [];
}
```

In this case, the `$request` parameter has no data, it's empty. Then getting the request from container solves the issue.

PR #681, opened 2017-11-27
Declaration of App\Console\Kernel::schedule should be compatible with Laravel\Lumen\Console\Kernel::schedule

ankitpokhrel:


taylorotwell:
The change you make here has nothing to do with the error you are receiving.

ankitpokhrel:
Ok the title is misleading but [Illuminate\Console\Scheduling\Schedule](https://github.com/illuminate/console/blob/master/Scheduling/Schedule.php) doesn't accept any params. Looks like It was removed long ago.

PR #690, opened 2017-12-16
Clean elses

carusogabriel:
I've removed the `else`s when we have already returned something :put_litter_in_its_place:


src/helpers.php
```diff
@@ -204,9 +204,9 @@ function factory()
             return $factory->of($arguments[0], $arguments[1])->times(isset($arguments[2]) ? $arguments[2] : null);
         } elseif (isset($arguments[1])) {
             return $factory->of($arguments[0])->times($arguments[1]);
-        } else {
-            return $factory->of($arguments[0]);
         }
+
+        return $factory->of($arguments[0]);
```
GrahamCampbell: 👎 This was done stylistically on purpose. If you were really cleaning the unused branches, you would also change every elseif to an if. 👎  on that too.

GrahamCampbell:
👎

PR #700, opened 2017-12-29
Provide dispatch_now along with dispatch function

maxklenk:
When the `dispatch()` function is available via `ProvidesConvenienceMethods` I would assume that I can also access the `dispatch_now()` function.


taylorotwell:
We never use snake case for class methods.

maxklenk:
I will create a new PR using `dispatchNow()` instead. 

I was confused because both functions were introduced in the upgrade guide:
```php
use Illuminate\Support\Facades\Bus;

$value = dispatch_now(new Job);

$value = Bus::dispatchNow(new Job);
```

PR #703, opened 2018-01-02
PHPDoc update for thrown Exception

texastoast33:
Simple update to PHPDoc to account for thrown Exception


taylorotwell:
Doesn't follow formatting of rest of framework.

Issue #607, opened 2017-04-19
Service Provider Boot method called before all Register methods called


Nais777:
- Lumen Version: 5.4
- PHP Version: 7.1
- Database Driver & Version: NA

### Description:

According to the **[documentation](https://lumen.laravel.com/docs/5.4/providers#writing-service-providers)** for the `boot()` method in Service Providers: **This method is called after all other service providers have been registered** but I am finding this to not be the case. Instead, the `boot()` method is called immediately after the `register()` method for a given Service Provider.

### Steps To Reproduce:

Create a new `Lumen` app: `Lumen new test-app`.

Enable the `AppServiceProvider` and `AuthServiceProvider` in `bootstrap\app.php`.

In each Service Provider, put `$this->app['log']->info("Booting AppServiceProvider");` in each `boot()` and `register()` methods, replacing `Booting` with `Registering` and `AppServiceProvider` with `AuthServiceProvider` where applicable. Alternativly, you could just write `Booting` or `Registering`.

Run `php artisan list`.

Look in `storage/logs/lumen.log` and you should see something like this:

~~~
[2017-04-19 22:04:50] lumen.INFO: Registering AppServiceProvider  
[2017-04-19 22:04:50] lumen.INFO: Booting AppServiceProvider  
[2017-04-19 22:04:50] lumen.INFO: Registering AuthServiceProvider  
[2017-04-19 22:04:50] lumen.INFO: Booting AuthServiceProvider  
~~~

Attached is a lumen application with the changes already made.

[test-lumen.zip](https://github.com/laravel/lumen-framework/files/938849/test-lumen.zip)

hongaar:
Just ran into the same problem. Seems like this has already been discussed before in https://github.com/laravel/lumen-framework/issues/399. However, this issue was closed early without a concrete answer or solution.

GrahamCampbell:
The docs are incorrect. Registering a service provider at runtime will register and boot it, immediately. The docs are referring to registering a service provider at config time, before the app has started.

GrahamCampbell:
I hope this distinction makes sense.

laurencei:
@GrahamCampbell - is there a way to register a service provider at config time in Lumen, since there is only a `.env` file? Or is the only way to do it at runtime?

hackel:
[...]

PR #723, opened 2018-02-11
Application boot

ragboyjr:
This PR introduces bootstrapping behavior consistent with Laravel framework. I really enjoy the simplicity of Lumen, but one of the first things I noticed is that the `boot` method on service providers is invoked immediately instead of after all of the service providers have been registered like in Laravel.

I tried to copy most of the code from the Illuminate\Foundation code as much as possible to keep it consistent (and have the most chance of getting merged). This is definitely a breaking change since I've modified the structure of a protected variable, and also changed the functionality of the booting.

Anyways, thanks for all of your hard work on these projects, hopefully this makes Lumen just a bit nicer to work with :).


ragboyjr:
This also addresses #607

ragboyjr:
Thanks 🎉

PR #724, opened 2018-02-11
App Level Response Builder and Error Formatter

ragboyjr:
If you want to configure the Routing response builder or error formatter, you need to call the `buildResponseUsing` on both the Closure and Controller class to affect both styles of controllers.

```php
$responseBuilder = function() {};

Laravel\Lumen\Routing\Closure::buildResponseUsing($validation_response_builder);
Laravel\Lumen\Routing\Controller::buildResponseUsing($validation_response_builder);
```

Or I just found while I was making this PR, you can actually call the trait's static method as well:

```php
Laravel\Lumen\Routing\ProvidesConvenienceMethods::buildResponseUsing($validation_response_builder);
```

However, this functionality isn't documented in PHP and also isn't very intuitive.

The change this PR makes is that you set the `buildResponseUsing` and `formatErrorsUsing` on the lumen application and the `ProvidesConvenienceMethods` reads from the app instead of static variables.

```php
$app = new Lumen\Application();
$app->buildResponseUsing(function() {});
$app->formatErrorsUsing(function() {});

// instead of

Lumen\Routing\Closure::buildResponseUsing(function() {});
Lumen\Routing\Controller::buildResponseUsing(function() {});
```

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -576,6 +576,32 @@ class Application extends Container
     }
 
     /**
+     * Define a response builder for building the failed validation responses.
+     *
+     * @param  callable  $fn
+     * @return $this
+     */
+    public function buildResponseUsing(callable $fn)
+    {
+        $this->instance('routing.responseBuilder', $fn);
+
+        return $this;
+    }
+
+    /**
+     * Define a error formatter for formatting the validation errors.
+     *
+     * @param  callable  $fn
+     * @return $this
+     */
+    public function formatErrorsUsing(callable $fn)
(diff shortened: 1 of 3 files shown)


taylorotwell:
No plans to add this.

ragboyjr:
Interesting, I’m not tied to my specific implementation, but I think the current way of setting a responseBuilder/errorFormatter is hokey.

If you have any other ideas, I wouldn’tind doing some of the leg work...

PR #725, opened 2018-02-14
Add default client in redis configuration

nicolas-wallerand:
Hi, is it possible to add the default client to redis in the default configuration. for more visibility


nicolas-wallerand:
As https://github.com/laravel/laravel/blob/master/config/database.php

or `'client' => env('REDIS_CLIENT', 'predis')`

PR #734, opened 2018-03-02
Updates cron expression to match Laravel

sixlive:
Updates expression package from `mtdowling/cron-expression:~1.0` to
`dragonmantank/cron-expression:~2.0` to match `laravel/framework`.

Tests pass but I didn't look to see if there was any coverage on the CRON expressions.


GrahamCampbell:
Does this need updating in the sub-packages?

Issue #667, opened 2017-10-13
Empty Request object in Resource when testing


jgrossi:
- Lumen Version: 5.5.1
- PHP Version: 7.1.8 
- Database Driver & Version: MariaDB 10.2.9

### Description:
When testing and making a request to a specific route, inside the Controller action, using the method injection for Request, I get all information I need:

```php
public function login(Request $request) // $request OK
```

But inside the `Resource` object, in all methods I receive it as a parameter, like `toArray($request)` or `with($request)`, this `$request` object is empty, no route information, no parameters, no path, nothing.

Then I have to get the request again from the service container:

```php
public function with($request) // empty
{
    $request = app(Request::class); // correct

    return $request->is('auth/login') && $this->jwtToken ?
        ['token' => $this->jwtToken] : [];
}
```

### Steps To Reproduce:

1. Create a test case with a call to an endpoint like `/foo`.

```php
$router->get('/foo', function () {
    $resource = App\User::find(1); // or any other model instance

    return new App\Http\Resource\Foo($resource);
});
```

Test case:

```php
public function test_resource_empty_request_object()
{
    $this->json('GET', '/foo');
}
```

2. Make the response a `Response` instance, like `FooResource` for example.

```php
namespace App\Http\Resource;

use Illuminate\Http\Request;
use Illuminate\Http\Resources\Json\Resource;

class Foo extends Resource
{
    /**
     * @param Request $request
     * @return array
     */
    public function toArray($request): array
    {
        $route = $request->route(); // empty
        $route = app(Request::class)->route(); // valid

        return $this->resource->toArray();
    }
}
```

3. Inside the `toArray($request)` method, check that the `$request` is empty, but `app(Request::class)` is not.

### Suggestion

In `Laravel\Lumen\Concerns\RoutesRequests::prepareResponse($response)` method, the request object is empty here (`Request::capture()`), but `app(Request::class)` is not:

```php
$response = $response->toResponse(Request::capture());
```

The `SymfonyRequest::createFromGlobals()` is always returning empty data, then the returned request is always empty.

If I change from `Request::capture()` to `app(Request::class)` it works, the `$request` in `Resource::toArray($request)` is not empty anymore.

```php
$response = $response->toResponse(app(Request::class));
```

### Related Issues

I think this "binding" issue is causing others issues with URL when testing. 

[...]

PR #739, opened 2018-03-11
[5.6] Add console request for url generation

### LINKED ISSUE #513: Generating URLs for named routes on commands is missing hostname and port
Hi,

Seems like the route helper is not working properly on commands. I would like to generate a URL to a named route with parameters but it seems to be always missing the hostname and port. It works fine on controllers and route closures but not on commands.

routes.php which returns http://localhost:12345/user/1
```
<?php
$app->get('user/{id}', ['as' => 'user', function ($id) {
    return route('user', ['id' => 1]);
}]);
```

Same route helper on the command:

```
/**
 * Execute the console command.
 *
 * @return mixed
 */
public function handle()
{
    $url = route('user', ['id' => 1]);
    $this->info($url);
}
```

```
vagrant@homestead:~/Code/lumen_route_helper$ php artisan test:url
http://:/user/1
```

gazben:
In a console environment there was no Request object. This caused problems with url generation during testing, and artisan commands.
Hopefully fixes #513 and #667.

### DIFF
--- a/src/Console/Kernel.php
+++ b/src/Console/Kernel.php
@@ -5,6 +5,7 @@ namespace Laravel\Lumen\Console;
 use Exception;
 use Throwable;
 use RuntimeException;
+use Illuminate\Http\Request;
 use Laravel\Lumen\Application;
 use Illuminate\Console\Scheduling\Schedule;
 use Illuminate\Console\Application as Artisan;
@@ -52,8 +53,8 @@ class Kernel implements KernelContract
     {
         $this->app = $app;
 
+        $this->setRequestForConsole($this->app);
         $this->app->prepareForConsoleCommand($this->aliases);
-
         $this->defineConsoleSchedule();
     }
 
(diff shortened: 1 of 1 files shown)


solofeed:
The issue with empty URL still exists. #513

gazben:
@solofeed In a clean lumen install from dev-master?
For reproduction do I need to install anything else?

solofeed:
Just create project via composer create and tested.
`dd(route('profile'))`
Output `string(16) "http://:/profile"`

gazben:
@solofeed 
Please check the vendor/laravel/lumen-framework/src/Console/Kernel.php constructor. Does it have a setRequestForConsole call?
If I clone the repo from dev-master the fix is not there yet. It's only available on the 5.6 branch.

solofeed:
@gazben 
I tested on laravel/lumen (v5.6.0), so after update to 5.6.2 all right work.
Thank you.

solofeed:
@gazben Sorry, but I am forget about force route in createApplication method.
`$app->make('url')->forceRootUrl(env('APP_URL', 'http://localhost/'));`
So without this, it will again not working.

gazben:
@solofeed The question still stands. Do you see the setRequestForConsole function in vendor/laravel/lumen-framework/src/Console/Kernel.php constructor?

solofeed:
@gazben Yes, i see

natepage:
Hi all,
The change made in this PR is actually breaking our unit tests by overriding the request so we're not able to authenticate any request anymore. 

Please have a look at the PR we've created to fix it #747 

Thank you

PR #747, opened 2018-03-26
[5.6] Console Kernel: bind request into container only if not already bound

natepage:
The change made into #739 bind request into the container regardless if a request has been already bound or not which has for result to give an empty request to my `viaRequest` closure in `AuthServiceProvider` so I'm not able to authenticate any user from the request anymore.

This fix bind the request into the container only if no request has been bound before.

### DIFF
--- a/src/Console/Kernel.php
+++ b/src/Console/Kernel.php
@@ -53,7 +53,7 @@ class Kernel implements KernelContract
     {
         $this->app = $app;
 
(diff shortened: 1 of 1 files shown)


iocaste:
I have same problem. After updating to lumen 5.6.2 my console tests fails. If I add this if statement into Console.php problem is solved.

PR #748, opened 2018-03-27
Test a job should not be fired by the application

anam-hossain:
The pull request includes a new test method which will verify that a job is not expecting by the application. The framework already includes a test method (`expectsJobs($jobs)`) to verify the job is being fired or not but nothing for the testing opposite.


taylorotwell:
Use fakes.

Issue #788, opened 2018-07-05
Multiple Lumen request instances in the container during integration test execution.


rihardsgrig:
- Lumen Version: 5.6.3
- PHP Version: 7.2

### Description:
I have a middleware that modifies request [ServerRequestInterface](https://github.com/php-fig/http-message/blob/master/src/ServerRequestInterface.php#L244) using method `withAttribute`. After the request changes, middleware sets the modified request into the container using the container's `instance` method, like `->instance(Request::class, MODIFIED_REQUEST_INSTANCE)`

If I'm launching my integration test, values stored in the request get lost. I see in the [MakesHttpRequests class](https://github.com/laravel/lumen-framework/blob/5.6/src/Testing/Concerns/MakesHttpRequests.php#L338), which is used for the testing, that request is set with the line `$this->app['request'] = Request::createFromBase($symfonyRequest);` which means that during the test execution there are multiple instances of the request.

### Steps To Reproduce:
- modify request using any `withXXXX` method from the ServerRequestInterface,
- set modified request into the container,
- launch integration test,
- see if expected values are present.

driesvints:
Isn't the solution just to use `->instance('request', MODIFIED_REQUEST_INSTANCE)` instead because the current request is registered like that in the container?

Closing this for now because this issue is a couple of months old already but feel free to reply if the issue persists.

PR #789, opened 2018-07-05
[5.6] Fix multiple request instances in the container during integration test

rihardsgrig:
This PR resolves [#788](https://github.com/laravel/lumen-framework/issues/788)

PR resolves the issue when the request instance that is modified by middleware is not the same request instance during the integration test.

### DIFF
--- a/src/Testing/Concerns/MakesHttpRequests.php
+++ b/src/Testing/Concerns/MakesHttpRequests.php
@@ -335,10 +335,10 @@ trait MakesHttpRequests
             $cookies, $files, $server, $content
         );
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
This seems like a breaking change?

rihardsgrig:
All the framework tests are green. And I've tested this on the multiple projects I have integration tests.

Maybe there is a different solution to the problem I described in the issue that is referenced?

crynobone:
Did you test this with project that extends `Request` instance such as `dingo/api`?

rihardsgrig:
no I'm not

taylorotwell:
Closing for now until more testing is done.

Issue #746, opened 2018-03-25
Lumen 5.4 - 5.6: Call to undefined method Laravel\Lumen\Application::routesAreCached()


GTCrais:
- Lumen Version: 5.4 - 5.6
- PHP Version: 7.1.10

### Description (and steps to reproduce):
You can call `$this->loadRoutesFrom()` within a `ServiceProvider`, which then calls `$this->app->routesAreCached()`, but that method does not exist.

dunhamjared:
Verified issue in v5.6.3

dunhamjared:
A quick dirty hack is to overwrite the function it's trying to call in the service provider. 

So in your `class AppServiceProviderExample extends ServiceProvider` class, add:

```php
//@TODO Remove me when a fix is implemented
public function loadRoutesFrom($location){
    return $location;
}
````

It's trying to call the `routesAreCached` function from `Illuminate\Foundation\Application` but `$this->app` is in `Laravel\Lumen\Application`.

**Comment by 4levels** [comment]
Hi Lumen team,

Any updates on this?  I'm also facing the exact same issue and all be it that the proposed fix works, this doesn't work for libraries that don't extend the AppServiceProvider (but use the Lumen/Support/ServiceProvider directly).

Kind regards,

**Comment by JacksonYara** [comment]
I don't think route caching is supported in Lumen. Maybe, this could be solved by adding the `routesAreCached` method to the `Laravel\Lumen\Application`, which returns false.

```PHP
/**
 * Determine if the application routes are cached.
 *
 * @return bool
 */
public function routesAreCached()
{
    return false;
}
```

driesvints:
This is fixed in Lumen 5.7

PR #792, opened 2018-07-20
Add the routesAreCached method, so Lumen won't throw a "Call to undef…

JacksonIV:
I made this pull request to fix the "Call to undefined method" regarding the `routesAreCached` method. It seems like route caching is not supported in Lumen, that's why i made it return false on default.

#746

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -820,6 +820,16 @@ class Application extends Container
     }
 
     /**
+     * Determine if the application routes are cached.
+     *
+     * @return bool
+     */
+    public function routesAreCached()
+    {
+        return false;
+    }
+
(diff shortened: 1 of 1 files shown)


kevmt:
This seems not carried over to latest lumen, was this removed? lumen-framework v5.6.4, this triggers errors to @nWidart package. Is composer using old or i am requiring the lumen-framework wrong?

driesvints:
@kevmt Lumen needs a new patch release for it. It's already committed just not released yet.

driesvints:
@kevmt alternatively you can upgrade Lumen to 5.7 where the issue is fixed.

Issue #685, opened 2017-12-05
[Routing] Call to a member function parameter() on array


abellion:
```php
$request->parameter;
// OR
$request->route('parameter');
// OR
$request->route()->getParameter('parameter');
```
Throws `Call to a member function parameter() on array`.

Bumping because all of the other issues are still alive :

https://github.com/laravel/lumen-framework/issues/291
https://github.com/laravel/lumen-framework/issues/650
https://github.com/laravel/lumen-framework/issues/484
https://github.com/laravel/lumen-framework/issues/119
...

hakuno:
Do

`list($found, $routeInfo, $params) = $request->route() ?: [false, [], []];`

Then

`dd($params);`

Is the expected thing on there?

yuloh:
I proposed a fix for this but it was rejected: https://github.com/laravel/internals/issues/71

Suggested workaround: https://github.com/laravel/lumen-framework/issues/484#issuecomment-300170119

Alternatively you can modify the calling code to expect an array:

```php
$param = $request->route()['parameter'] ?? null;
```

xxRockOnxx:
This issue still exists.

codefocus:
Confirming that this issue still exists.

`Call to a member function parameter() on array in vendor/illuminate/http/Request.php:458`

In my case, this was triggered by calling `$request->json->all()` from a Lumen Controller.

xxRockOnxx:
I just read from an issue that Taylor does not consider it as a bug and says Laravel and Lumen is not 100% compatible so it seems like this will not be fixed.

MrPatrickHenry:
Just to confirm this same issue: "/management/stream/{parameter}" in the uri following lumen documentation. i get this error

```
1/1) FatalThrowableError Call to a member function parameter() on array
--
in Request.php line 501
at Request->route('parameter')in Request.php line 665
```

I was breaking things out in to microservice under the impression that lumen can be migrated in to laravel quite easily.

abellion:
Thanks @JacksonIV

JacksonIV:
@abellion Anytime!

lloricode:
this is already fixed, but when running on test is still same issue

loordy:
> 

> hi @abellion
> this is already fixed, but when running on test is still same issue
> 
> Laravel Framework Lumen (5.7.6)
> PHPUnit 7.4.4
> 
> ```
[...]

Issue #650, opened 2017-09-11
5.5 setRouteResolver no support in Illuminate\Http\Request


Except10n:
- Lumen Version: 5.5.0
- PHP Version:7.2
- Database Driver & Version: NAN

### Description:
$request['xxxxxx']  throw an exception, 

> Call to a member function parameter() on array

Lumen set the resolver will return one array
```
$this->currentRoute = $routeInfo;

$this['request']->setRouteResolver(function () {
    return $this->currentRoute;
});

```

Illuminate\Http\Request > __get()  > route()  call parameter on array
```
$route = call_user_func($this->getRouteResolver());

        if (is_null($route) || is_null($param)) {
            return $route;
        }

        return $route->parameter($param);
```

driesvints:
Closing this issue because it's already solved, old or not relevant anymore. Feel free to reply if you're still experiencing this issue.

Issue #484, opened 2016-09-16
Request:route() method throws exception


ObliviousHarmony:
Hello,

This Request method throws an exception:

``` php
    /**
     * Get the route handling the request.
     *
     * @param string|null $param
     *
     * @return \Illuminate\Routing\Route|object|string
     */
    public function route($param = null)
    {
        $route = call_user_func($this->getRouteResolver());

        if (is_null($route) || is_null($param)) {
            return $route;
        } else {
            return $route->parameter($param);
        }
    }
```

In Lumen, the route resolver returns an array, and so the function call fails. I would have submitted a pull request but I'm not sure how you might want this problem solved, since the Request class is part of Laravel.

As a slight addendum, I discovered this while creating a Lumen friendly FormRequest. Is this something you guys might be interested in a pull request for? It has its own service provider, so it's an optional thing that might be nice for other developers.

yuloh:
I opened a proposal [to create a common route interface](https://github.com/laravel/internals/issues/71) back in April but I haven't received a response.  I may just submit a PR and see what happens.

abellion:
Same issue here, any new on that @taylorotwell  ? Did you guys found a workaround ?

yuloh:
If you check the the issue I linked it doesn't sound like this is going to be fixed :(

You can probably workaround it with a route middleware that replaces the array with an object, or by extending the service container binding.  I don't think lumen fires an event when the route is matched.


```php
// in a service provider somewhere
$this->app->resolving(Request::class, function ($request) {
    if (!is_array($request->route())) {
        return;
    }
    $route = Route::fromArray($request->route());
    $request->setRouteResolver(function () use ($route) {
        return $route;
    });
});


class Route
{
    private $parameters;

    public function __construct(array $parameters)
    {
        $this->parameters = $parameters;
    }

    public static function fromArray(array $route)
    {
        return new self($route[2]); // 0 is the Dispatcher::FOUND code, 1 is the action, 2 is the params.
    }

    public function parameter($name, $default = null)
    {
        return array_get($this->parameters, $name, $default);
    }
}
```

JacksonIV:
This is fixed in 5.7.

driesvints:
@JacksonIV thanks for checking 👍

wajdijurry:
If you're using lumen (5.7 and above), I figured out a good solution to solve this:
[...]

Issue #291, opened 2015-12-17
Can't call parameter on routeResolver array


guidsen:
Trying to read route parameters from my Middleware class with `$request->route(‘id’)`.  This gives me the error that it can’t call `parameter()` on an array. When I look at the source code, the `$this->getRouteResolver()` returns an array. Is this a bug?

```
public function route($param = null)
{
  $route = call_user_func($this->getRouteResolver());

  if (is_null($route) || is_null($param)) {
    return $route;
  } else {
    return $route->parameter($param);
  }
}
```

GrahamCampbell:
Yes, I guess it is a bug.

GrahamCampbell:
It's not a bug that it returns an array, but then the fact we try to treat it as an object is a bug.

guidsen:
Currently using `$routeParams = $request->route()[2];` and `$routeParams['paramName']` to retrieve a route parameter from the request. I'm not sure if the index `[2]` will be always the same tho. Seems it does.

irazasyed:
This issue has been around for quite sometime (Since Lumen's release actually!) and there has never been a fix to it till date.

Ref: #198 #119

I had even sent a PR because i was having this very same issue back then (Though it's just a helper which never made it to the core): #52

I actually ended up writing a new package with that helper [here](https://github.com/irazasyed/larasupport/blob/master/src/helpers.php#L3-L19).

I'm once again having the same issue, this time with a third-party package which uses `$request->route()` to access the param.

@GrahamCampbell Will there ever be a fix for this :?

GrahamCampbell:
We're open to PRs.

irazasyed:
Alright, I'll create a new PR with solution soon then or if @guidsen is doing, that should also be good too.

irazasyed:
Sent a PR: https://github.com/laravel/framework/pull/11873

Please review. Tested both in Lumen and Laravel, Works good.

irazasyed:
P.S We could also add support for the default value, isn't it? Since the `parameter()` method in Laravel Route class already supports it but it isn't added in the request's route method.

irazasyed:
@taylorotwell @GrahamCampbell Since the PR in Laravel was rejected because it should extend the request class in Lumen, Should i go ahead and do that or wait for such thing happening? Lmk.

SPRING-Z:
why longtime?

SPRING-Z:
@irazasyed @taylorotwell @GrahamCampbell

[...]

PR #793, opened 2018-07-23
[5.6] Feature/fix route parameter

JacksonIV:
I made this pull request, to fix the `Call to a member function parameter() on array` error. This would occur when you called the `route` method on the `Illuminate\http\Request` class in Lumen. I tried to follow the advice in this pull request: #11873

#291
#650
#484
#119 
#685

### DIFF
--- a/src/Concerns/RoutesRequests.php
+++ b/src/Concerns/RoutesRequests.php
@@ -8,6 +8,7 @@ use Throwable;
 use FastRoute\Dispatcher;
 use Illuminate\Support\Str;
 use Illuminate\Http\Request;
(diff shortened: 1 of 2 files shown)


taylorotwell:
If we were going to do this I think the signature should match Laravel's?

JacksonIV:
@taylorotwell Thank you for responding. What do you mean exactly by ‘the signature’?

JacksonIV:
@taylorotwell I could encapsulate the `$this->currentRoute` into it's own `Route` class. That way i could make a `parameter` method. Laravel also has a `Route` class, right?

Issue #119, opened 2015-05-30
Request->route($param) doesnt work


leemason:
trying to access route parameters through the request object (ie in controller constructor) doesnt work.

calling route() without any arguments works fine and returns an array (which i assume comes from fast route) however trying to get a param returns fatal error "parameter method doesnt exist on none object)

agreed fastroute isnt going to be as feature full as the beefy symfony alternative, but this should either be fixed or documented as not allowed.

GrahamCampbell:
Could you provide a short code example please?

GrahamCampbell:
Also, note that we only support lumen 5.1, not 5.0.

yaede7:
I also have this issue. Looking at the code it seems the Request->route() returns an array and not an instance of \Illuminate\Routing\Route as documented

This is the route function from Illuminate\Http\Request:
```php
    /**
     * Get the route handling the request.
     *
     * @return \Illuminate\Routing\Route|null
     */
    public function route()
    {
        if (func_num_args() == 1) {
            return $this->route()->parameter(func_get_arg(0));
        } else {
            return call_user_func($this->getRouteResolver());
        }
    }
```
 To access the route parameters I have done $request->route()[2][$param];

Not sure if this is suppose the return an instance of \Illuminate\Routing\Route or an array though.

GrahamCampbell:
>  returns an array

Does it?

GrahamCampbell:
This issue is unrelated to Lumen. I'll have a look at that laravel function.

GrahamCampbell:
Check out https://github.com/laravel/framework/pull/9417

alexw23:
:+1: on this issue

alexw23:
All the documentation points to `$request->route()` returning an object, but it returns an array.

```
array:3 [
  0 => 1
  1 => array:13 [
    "middleware" => array:2 [
      0 => "api.auth"
      ...
    ]
    ...
  ]
  2 => array:1 [
    "customerId" => "3"
  ]
]
```

`$request->route()->setParameter('User', $user);` returns `Call to a member function setParameter() on array`

GrahamCampbell:
Yes, the docs are just wrong.

GrahamCampbell:
The array replaces a route object.

crishoj:
Also note that when a request is invoked from a `Laravel\Lumen\Testing\TestCase` using the `$this->post(...)` syntax, `$request->route` suddenly returns `NULL`.

[...]

PR #799, opened 2018-08-10
[5.6] Fix missing parameter function on Request.

JacksonIV:
This is a follow up on my closed Pull Request #793. I made some adjustments and i've added a test.

> I made this pull request, to fix the Call to a member function parameter() on array error. This would occur when you called the route method on the Illuminate\http\Request class in Lumen. I tried to follow the advice in this pull request: #11873

#291
#650
#484
#119
#685


taylorotwell:

Thanks!

PR #803, opened 2018-09-09
[5.7] Extension of Illuminate's Request class within Lumen, in order to fix an incompatibility between Laravel and Lumen.

JacksonIV:
This pull request aims to fix an incompatibility between Laravel and Lumen's route resolvers. Since Lumen uses a different router, namely "fast-route", there is a difference between the two route resolvers. The difference causes an `Call to a member function parameter() on array` exception while trying to call the `route` method with a parameter in Illuminate's Request class.

The `route` method, when a parameter has been passed, can be used to extract the GET value of the current route. It will try to do so by calling the `parameter` method on the route resolver. This works fine within Laravel, because the route resolver returns an object. Within Lumen, however, the route resolver will return an array. Hence the exception.

In order to fix this issue, i created a custom Request class within Lumen. This class extends the original Illuminate Request class and overwrites the `route` method. The new `route` method will try to extract the GET parameters from the route resolver's array, instead of calling a method on the array.

When Lumen tries to parse the incoming request, it will use the new Request class instead of the original one. It will automatically bind the new Request class to original Illuminate Request class. This way, i tried to make this change as backwards compatible as possible.

Within Lumen's test i had to replace the original Request class's use statement with the new one. That's because Illuminate's Request class was called directly. However, since the new Request class extended the original Request class it should still pass the tests. In order to test this a little more i made an assertion which whether the new Request class is an instance of the original Illuminate's Request class.

In addition to that, i modified the `testRequestWithParameters` test with two new assertions. Those assertions test whether the new `route` method successfully extracts the GET parameters from the route.

I hope my explanation is thorough enough and that my fix is sufficient. I referenced a few issues, if you'd like to find out some more about this issue.

#291
#650
#484
#119
#685

### DIFF
--- a/src/Concerns/RoutesRequests.php
+++ b/src/Concerns/RoutesRequests.php
@@ -11,6 +11,7 @@ use Illuminate\Http\Request;
 use Illuminate\Http\Response;
 use Laravel\Lumen\Routing\Pipeline;
 use Illuminate\Contracts\Support\Responsable;
+use Laravel\Lumen\Http\Request as LumenRequest;
 use Laravel\Lumen\Routing\Closure as RoutingClosure;
 use Illuminate\Http\Exceptions\HttpResponseException;
 use Laravel\Lumen\Routing\Controller as LumenController;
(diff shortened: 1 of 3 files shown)


taylorotwell:
Why does the signature not match the Laravel request? It's missing the default parameter.

JacksonIV:
@taylorotwell I've updated the signature to match Laravel's. Something went wrong on my side.

Issue #808, opened 2018-09-14
[5.7] Tests using the be() and/or actingAs() helpers fail


stefpankov:
- Lumen Version: 5.7.0
- PHP Version: 7.2

### Description:
Tests using the be() and/or actingAs() helpers but not making an actual request fail with the following message:

> InvalidArgumentException: Auth driver [api] for guard [api] is not defined.

After investigating the issue, I concluded that it's due to the newest changes introduced in the Application class which require the application to be explicitly booted using the [boot()](https://github.com/laravel/lumen-framework/compare/5.6...5.7#diff-c9248b3167fc44af085b81db2e292837L204) method which is only called from within the [dispatch method](https://github.com/laravel/lumen-framework/compare/5.6...5.7#diff-145186519daafc23b06eb900961a02cfR159) of the RoutesRequests trait.

I would submit a patch for this but I'm not sure on where to call the boot method for a testing scenario. It seems like a good place do to so would be the refreshApplication() method of the TestCase class.

### Steps To Reproduce:
1. Create a new test.
```php
<?php

use Illuminate\Auth\GenericUser;
use Test\TestCase;

class AuthTest extends TestCase
{
    public function setUp()
    {
        parent::setUp();

        $user = new GenericUser([
            'id' => 1
        ]);

        $this->be($user);
    }

    public function testAuth()
    {
         $this->assertTrue(true);
    }
}
```
2. Run `phpunit`

georgeboot:
Did you also encounter #806? I had that issue first, and then after my workaround, this one happened.

stefpankov:
I don't have any tests involving models, I would do a quick one to check but am not currently at my desk, if you could, go to the lumen TestCase class and in the refreshApplication method right under [here](https://github.com/laravel/lumen-framework/blob/5.7/src/Testing/TestCase.php#L56) add a call to `$this->app->boot();` and see if that works for both cases and without your workaround.

georgeboot:
@stefpankov I can confirm that workaround indeed works.

JacksonIV:
I made a pull request to fix this issue.

georgeboot:
Fixed in #809

Issue #806, opened 2018-09-14
[5.7] Eloquent Models have no ConnectionResolver


georgeboot:
- Lumen Version: 5.7.0
- PHP Version: 7.2.9 cli

### Description:
When running tests in Lumen 5.7, the Eloquent models do not get an `ConnectionResolver`, and thus no connection.

This produces the following error:
```
1) Tests\Feature\AccessLoggerTest::it_logs_requests_with_a_user
Error: Call to a member function connection() on null

/path/to/code/vendor/illuminate/database/Eloquent/Model.php:1234
/path/to/code/vendor/illuminate/database/Eloquent/Model.php:1200
/path/to/code/vendor/illuminate/database/Eloquent/Model.php:1030
/path/to/code/vendor/illuminate/database/Eloquent/Model.php:945
/path/to/code/vendor/illuminate/database/Eloquent/Model.php:983
/path/to/code/vendor/illuminate/database/Eloquent/FactoryBuilder.php:203
/path/to/code/vendor/illuminate/support/Collection.php:407
/path/to/code/vendor/illuminate/database/Eloquent/FactoryBuilder.php:207
/path/to/code/vendor/illuminate/database/Eloquent/FactoryBuilder.php:181
/path/to/code/tests/Feature/AccessLoggerTest.php:42
```

### Steps To Reproduce:

1. Inside a test, do an eloquent call. Mine was `$user = factory(ApiUser::class)->create();`
2. Make sure your `phpunit.xml` set's your env variables correctly (I copied mine from the master of this repo). Make sure both `APP_ENV` and `DB_CONNECTION` are set to `testing`.
3. Make sure you enable `$app->withEloquent()` in `bootstrap/app.php`. Enabling or disabling `$app->withFacades()` has no effect, so you can leave that one disabled.
3. Run `vendor/bin/phpunit`
4. Experience the above error

### Workaround
Override your Eloquent models constructor and add the following:
```php
public function __construct(array $attributes = [])
{
    $resolver = app('Illuminate\Database\ConnectionResolverInterface');
    
    $this->setConnectionResolver($resolver);
    
    parent::__construct($attributes);
}
```

### Updated workaround
Add the following method to your `BaseTest`. You don't need to change the Models.
```php
public function refreshApplication()
{
    parent::refreshApplication();

    $this->app->boot();
}
```

ISaidHey:
Also experiencing this issue and the updated workaround works for me.

flonology:
[...]

PR #809, opened 2018-09-17
[5.7] Boot the application in the testing environment.

JacksonIV:
Due to some new changes in Lumen, the application only boots within the `RoutesRequests` class. This pull requests ensures that the app will also boot when it's in testing mode.

#808
#806


taylorotwell:
Do we know what change actually caused this?

JacksonIV:
@taylorotwell I think this commit causes the issue:

eceb66305bcd22bec9e8390c6c1978f666f429b3

PR #813, opened 2018-09-27
Include Symfony VarDumper component

klemenb:
After https://github.com/illuminate/support/commit/98cf3233c2e4a54c9aeba661e0ec40ccc3285dc0 the `dd` helper no longer exists in Lumen.

This pull requests adds the `symfony/var-dumper` to the framework's `composer.json` and brings the helper back.


GrahamCampbell:
Why not add this to suggested dependencies instead?

PR #821, opened 2018-10-18
[5.8] Removed obsolete catch

martinssipenko:
All exceptions implement Throwable interface thus given catch statement is obsolete.


GrahamCampbell:
Thanks. :)

PR #825, opened 2018-10-19
[5.7] Set the application url when running tests.

JacksonIV:
When running tests, the application will boot through the console and will therefore not have a base url. This issue was fixed for regular commands in the kernel. The kernel, however, will never be instantiated when running tests, because it will boot the application in the regular way.

This PR sest the root url of the url generator to the APP_KEY environment value when Lumen boots in the testing environment.

#513

### DIFF
--- a/src/Testing/TestCase.php
+++ b/src/Testing/TestCase.php
@@ -55,6 +55,8 @@ abstract class TestCase extends BaseTestCase
 
         $this->app = $this->createApplication();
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Surely we should head to config for this, instead of env. That variable could be named anything at all.

GrahamCampbell:
@taylorotwell

PR #827, opened 2018-10-20
[5.7] Don't unnecessarily unset available bindings

yuloh:
This was a workaround and is not necessary since 56277d4 was merged.


GrahamCampbell:
Thanks.

PR #828, opened 2018-10-20
[5.7] Allow generating URLs with optional parameters

### LINKED ISSUE #539: Helper route() don't recognize optional params
Hi, friends.

I'm trying to use the route() helper as the example:

**NfeController@create**
`$urlParams = ['emitterId' => $emitter->id, 'status' => 'entradas', 'key' => $nfe->chave, 'format' => 'xml'];`
`$url = route('nfe', $urlParams);`

**routes.php**
`$app->group(['prefix' => 'emitter'], function($app){`
    `$app->group(['prefix' => '{emitterId}/nfe'], function($app){`
            `$app->get('{status}/{key}[/{format}]', ['as' => 'nfe', 'uses' => 'NfeController@show']);`
    `}`
`}`

but the function returns:
http://localhost:8000/emitter/36/nfe/entradas/29161207683985000146550010000000151000000151[/xml]

The helper **route** supports optional params?

P.S.: lumen 5.3; nikic/fast-route: 1.0.1

yuloh:
This PR makes it possible to generate routes for URLs with optional parameters.  Resolves #539.

To make this possible I added a second `preg_replace_callback` that replaces the optional part of the route.  I'm pretty sure it's not possible to do this with a single regex (Nikic said the same thing [here](https://github.com/nikic/FastRoute/issues/66#issuecomment-130395124)).

The regex is pretty dense so I will attempt to briefly summarize it.  You can view an example on [regexr](https://regexr.com/41jvr).

Unlike required parameters optional parameters might have non dynamic segments before and after, i.e. `/{id:\d+}/profile]`.  This is the reason capture groups 1 and 5 are added to the replacement for capture group 2.

It's also possible for an optional segment to not have any parameters, i.e. `/foo[bar]`.  In this case it's not clear if we should generate `/foobar` or `/foo`.  I went with the shorter URL.

We need to match this optional segment so we can remove it, but we don't want to match the square brackets inside a parameter, i.e. `/users/{id:[0-9]+}`.  Luckily you can only have an optional parameters in the trailing position so we can use `$` instead of a negative lookahead.

### DIFF
--- a/src/Routing/UrlGenerator.php
+++ b/src/Routing/UrlGenerator.php
@@ -257,6 +257,14 @@ class UrlGenerator
 
         $parameters = $this->formatParametersForUrl($parameters);
 
+        $uri = preg_replace_callback('/\[([^{]*)\{?(.*?)(:.*?)?(\{[0-9,]+\})?\}?([^}]*)\]$/', function ($m) use (&$parameters) {
+            if (! isset($parameters[$m[2]])) {
+                return '';
+            }
+
+            return $m[1].array_pull($parameters, $m[2]).$m[5];
+        }, $uri);
+
         $uri = preg_replace_callback('/\{(.*?)(:.*?)?(\{[0-9,]+\})?\}/', function ($m) use (&$parameters) {
             return isset($parameters[$m[1]]) ? array_pull($parameters, $m[1]) : $m[0];
         }, $uri);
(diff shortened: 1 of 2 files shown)


taylorotwell:
I don't want to maintain this regex.

yuloh:
To be clear the majority of the regex (everything between the curly braces) is the same as the required parameter regex below.  Would you prefer if that was pulled out into a constant and reused?

Alternatively we could use the route parser from fast route to do the bulk of the work like nikic recommends in the issue i linked above. I went with the regex to be consistent with the existing code but the route parser works fine too.

Let me know if there is a solution you would be happy with.

PR #831, opened 2018-10-25
[5.7] Update config files. Add services.php

chimit:
Update configs to make them more similar to Laravel's.

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -4,19 +4,6 @@ return [
 
     /*
     |--------------------------------------------------------------------------
-    | PDO Fetch Style
-    |--------------------------------------------------------------------------
-    |
-    | By default, database results will be returned as instances of the PHP
-    | stdClass object; however, you may desire to retrieve records in an
-    | array format for simplicity. Here you can tweak the fetch style.
-    |
-    */
-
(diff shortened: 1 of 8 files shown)


config/database.php
```diff
@@ -46,51 +33,50 @@
...
+            'port' => env('DB_PORT', 3306),
+            'database' => env('DB_DATABASE', 'forge'),
+            'username' => env('DB_USERNAME', 'forge'),
+            'password' => env('DB_PASSWORD', ''),
+            'unix_socket' => env('DB_SOCKET', ''),
+            'charset' => env('DB_CHARSET', 'utf8mb4'),
             'collation' => env('DB_COLLATION', 'utf8mb4_unicode_ci'),
-            'prefix'    => env('DB_PREFIX', ''),
-            'timezone'  => env('DB_TIMEZONE', '+00:00'),
-            'strict'    => env('DB_STRICT_MODE', false),
```
sksethi25: default value for strict mode is changed from false to true.Some users may face some db errors untill they figured out this

chimit: It was changed to conform to Laravel defaults.

PR #833, opened 2018-10-26
Database engine config fix

chimit:
The bug causes migration problems.


driesvints:
Can you please give some more info?

chimit:
In my [previous PR](https://github.com/laravel/lumen-framework/pull/831) I apparently, specified a wrong default value for `engine` while it [must be `null`](https://github.com/laravel/laravel/blob/master/config/database.php#L55) based on Laravel.

`false` causes SQL syntax error on migrations table creation.

PR #837, opened 2018-10-28
add InnoDB as default value

hamedbrd:
We could not run migration if we have no default value DB_ENGINE


taylorotwell:
You can do this on your own configuration.

hamedbrd:
But When i pulled repository and run migration i got error because of DB_ENGINE, so what is the solution??

driesvints:
Hey @hamedbrd. This is fixed with https://github.com/laravel/lumen-framework/pull/833 but we still need to tag a release. We'll probably do that soon.

PR #840, opened 2018-10-30
[5.7] Change application instance

hack3p:
Hi, i have a some problems with my custom `Application` which extends from base `lumen\Application` and setup in `bootsrap/app.php`.

```php
// Custom\Application.php with custom methods
class Application extends \Laravel\Lumen\Application 

// app.php as base setup
use Custom\Application;

$app = new Application(
    realpath(__DIR__ . '/../')
);

```
This fix solve my problems and has back compatibility.
 What do you think?

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -107,7 +107,7 @@ class Application extends Container
         static::setInstance($this);
 
         $this->instance('app', $this);
-        $this->instance('Laravel\Lumen\Application', $this);
+        $this->instance(static::class, $this);
 
         $this->instance('path', $this->path());
 


krenor:
This is **not** backwards compatible.
We overrode [the bootstrapping](https://github.com/laravel/lumen/blob/master/bootstrap/app.php#L22) with a customized Application class, too.  
After updating from `6.7.3` to `6.7.5`, where this is included, pretty much everything behaves wonky and most of our tests fail in the CI.  
Even a simple `php artisan tinker` results in `In Container.php line 945: Target [Illuminate\Contracts\Debug\ExceptionHandler] is not instantiable.`

The only way to get this running again is by aliasing it:
`$app->alias(\App\Application::class, \Laravel\Lumen\Application::class);`

JacksonIV:
@krenor could you make an issue? I can try to help you from there. 😬

krenor:
@JacksonIV I've created a PR reverting this change instead. Have a look: https://github.com/laravel/lumen-framework/pull/847

PR #848, opened 2018-11-13
[5.7] Fix overwriting the Application class

driesvints:
See https://github.com/laravel/lumen-framework/pull/840. Added a test to verify the correct behavior.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -107,7 +107,7 @@ class Application extends Container
         static::setInstance($this);
 
         $this->instance('app', $this);
-        $this->instance(static::class, $this);
+        $this->instance(self::class, $this);
 
         $this->instance('path', $this->path());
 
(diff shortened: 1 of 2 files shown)


crynobone:
But can Lumen resolve to `App\Application` when you try to use dependency injection by type-hinting `Laravel\Lumen\Application`?

crynobone:
I would say reverting the changes is still the best bet, anyone who want to extends Application should be able to alias it under `bootstrap.php`.

driesvints:
> But can Lumen resolve to App\Application when you try to use dependency injection by type-hinting Laravel\Lumen\Application?

@crynobone yes, that's what the tests proves. 

> I would say reverting the changes is still the best bet, anyone who want to extends Application should be able to alias it under bootstrap.php.

But with this PR you wouldn't have to do that. What makes you think we should revert this?

crynobone:
Sorry I got confuse with what this trying to revert to vs what #847.

driesvints:
No worries :)

Issue #849, opened 2018-11-15
V 5.7.5 breaks  Binding Resolution


j3rrey:
- Lumen Version: 5.7.5
- PHP Version: 7.2

### Description:
I've just done an upgrade from 5.5 to 5.6 to 5.7 until 5.7.4 everything works as expected, as soon as I pull in 5.7.5.
Php Artisan gives me "Target [Illuminate\Contracts\Debug\ExceptionHandler] is not instantiable."

Example:
```
php artisan

In Container.php line 945:

  Target [Illuminate\Contracts\Debug\ExceptionHandler] is not instantiable.
```

### Steps To Reproduce:

upgrade from 5.5 to 5.6 to 5.7 and fin ally from 5.7.4 to 5.7.5.

I'm checking all the code changes and what could possibly be causing this atm.

Maybe you guys have an idea ? 

Here is the stacktrace: 
```
#[2018-11-15 15:30:50] local.ERROR: Illuminate\Contracts\Container\BindingResolutionException: Target [Illuminate\Contracts\Debug\ExceptionHandler] is not instantiable. in /project/vendor/illuminate/container/Container.php:945
#Stack trace:
#0 project/vendor/illuminate/container/Container.php(785): Illuminate\Container\Container->notInstantiable('Illuminate\\Cont...')
#1 project/vendor/illuminate/container/Container.php(658): Illuminate\Container\Container->build('Illuminate\\Cont...')
#2 project/vendor/illuminate/container/Container.php(609): Illuminate\Container\Container->resolve('Illuminate\\Cont...', Array)
#3 project/vendor/laravel/lumen-framework/src/Application.php(260): Illuminate\Container\Container->make('Illuminate\\Cont...', Array)
#4 project/vendor/illuminate/container/Container.php(1222): Laravel\Lumen\Application->make('Illuminate\\Cont...')
#5 project/vendor/laravel/lumen-framework/src/Console/Kernel.php(223): Illuminate\Container\Container->offsetGet('Illuminate\\Cont...')
#6 project/vendor/laravel/lumen-framework/src/Console/Kernel.php(117): Laravel\Lumen\Console\Kernel->reportException(Object(Illuminate\Contracts\Container\BindingResolutionException))
#7 project/artisan(43): Laravel\Lumen\Console\Kernel->handle(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
[...]

PR #850, opened 2018-11-16
[5.7] Fixed situation with overridden base Application-class

mingalevme:
My Application-class (**App\Helpers\Laravel\Lumen\Application**) override base-class (**Laravel\Lumen\Application**). The overriding was made in the middle of project life so part of the code expects **Laravel\Lumen\Application** and part - **App\Helpers\Laravel\Lumen\Application**.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -108,6 +108,9 @@ class Application extends Container
 
         $this->instance('app', $this);
         $this->instance(self::class, $this);
+        if (self::class !== static::class) {
+            $this->instance(static::class, $this);
+        }
 
         $this->instance('path', $this->path());
 


GrahamCampbell:
What's your use case for extending the application class?

mingalevme:
```
public function isDownForMaintenance()
    {
        $filename = app()->storagePath() . '/lib/mfeed/down';
        return file_exists($filename);
    }
```
and

```
public function call($callback, array $parameters = [], $defaultMethod = null)
    {
        if (is_array($callback) && $callback[0] instanceof Command && $callback[1] === 'handle') {
            \Log::debug("Command is executing", [
                'command' => get_class($callback[0]),
            ]);
            $response = \Event::fire(new CommandExecuting($callback[0], $parameters), [], true);
            if ($response !== null) {
                \Log::debug("Execution of the command has been halted by a listener", [
                    'command' => get_class($callback[0]),
                    'response' => $response,
                ]);
                return null;
            }
        }

        $start = microtime(true);
        $result = parent::call($callback, $parameters, $defaultMethod);
        $time = microtime(true) - $start;

        if (is_array($callback) && $callback[0] instanceof \App\Jobs\BaseJob && $callback[1] === 'handle') {
            \Event::fire(new JobProcessed($callback[0]));
            \Log::debug("Job has been processed", [
                'job' => get_class($callback[0]),
                'time' => round($time, 3),
            ]);
        }

        if (is_array($callback) && $callback[0] instanceof Command && $callback[1] === 'handle') {
            \Event::fire(new CommandExecuted($callback[0], $parameters));
            \Log::debug("Command has been executed", [
                'command' => get_class($callback[0]),
            ]);
        }
        
        return $result;
    }
```

driesvints:
@mingalevme I don't understand why you'd use two separate `Application` classes. You either you the one provided by Lumen or overwrite it with your custom one.

Btw you `isDownForMaintenance` has an `app()` call which can be replaced by `$this`.

mingalevme:
I've fixed it by:
```$app->alias(App\Helpers\Laravel\Lumen\Application::class, \Laravel\Lumen\Application::class);```
but I think it may be useful for someone else.

**Comment by driesvints** [comment]
> I've fixed it by:
> `$app->alias(App\Helpers\Laravel\Lumen\Application::class, \Laravel\Lumen\Application::class);`
> but I think it may be useful for someone else.

You should only replace `$app = new Laravel\Lumen\Application(` in your `bootstrap/app.php` file with `$app = new App\Helpers\Laravel\Lumen\Application(` really.

**Comment by driesvints** [comment]
@mingalevme also: can you check if you're on 5.7.6?

```bash
php artisan version
```

mingalevme:
> @mingalevme I don't understand why you'd use two separate `Application` classes. You either you the one provided by Lumen or overwrite it with your custom one.
> 
I use only one Application-class - my custom, I added it not in the begging of the project's life, but when I added it I didn't replace all occurrences of ```Laravel\Lumen\Application```, and that worked in 5.6-. Now I've upgraded to 5.7 and that has been broken.

mingalevme:
> $app = new App\Helpers\Laravel\Lumen\Application(...)
> 
That I did when override the class.

mingalevme:
The biggest problem with it - that it take a lot of time to understand what exactly has been breaking after upgrading to 5.7.

mingalevme:
> @mingalevme also: can you check if you're on 5.7.6?
> 
> ```shell
> php artisan version
> ```

Lumen has no ```version``` command

driesvints:
@mingalevme if you upgrade to 5.7.6 this should be fixed.

driesvints:
See https://github.com/laravel/lumen-framework/issues/849

driesvints:
@mingalevme why would you expect the base class if you're overwriting it? The `bootstrap/app.php` has the only occurrence.

mingalevme:
Because there are legacy classes, I didn't want to touch them to didn't break something.

PR #853, opened 2018-11-27
[5.7] Remove unnecessary code

MilesChou:
The setRouteResolver() method is called when prepareRequest() method, so it's unnecessary to call in handleFoundRoute() method.

See https://github.com/laravel/lumen-framework/blob/v5.7.6/src/Application.php#L488-L490

### DIFF
--- a/src/Concerns/RoutesRequests.php
+++ b/src/Concerns/RoutesRequests.php
@@ -244,10 +244,6 @@ trait RoutesRequests
     {
         $this->currentRoute = $routeInfo;
 
-        $this['request']->setRouteResolver(function () {
-            return $this->currentRoute;
-        });
-
         $action = $routeInfo[1];
 
         // Pipe through route middleware...


GrahamCampbell:
Why was this added in https://github.com/laravel/lumen-framework/pull/381 then?

MilesChou:
At the #381 commit ea2008897be06f847ec888ea34ee9907da1608e4, `setRouteResolver()` will be called in the `prepareRequest()` and `parseIncomingRequest()` didn't call `prepareRequest()` when request is not from external.

> See the [link](https://github.com/laravel/lumen-framework/blob/ea2008897be06f847ec888ea34ee9907da1608e4/src/Concerns/RoutesRequests.php#L396-L406)

```php
protected function parseIncomingRequest($request)
{
    if ($request) {
        $this->instance(Request::class, $this->prepareRequest($request));
        $this->ranServiceBinders['registerRequestBindings'] = true;

        return [$request->getMethod(), $request->getPathInfo()];
    } else {
        // Here: didn't set Container with prepareRequest()
        return [$this->getMethod(), $this->getPathInfo()];
    }
}
```

And it's will be call at version [v5.7.6](https://github.com/laravel/lumen-framework/blob/v5.7.6/src/Concerns/RoutesRequests.php#L183-L192)

```php
protected function parseIncomingRequest($request)
{
    if (! $request) {
        $request = LumenRequest::capture();
    }

    // Definitely, the request with route resolver will be set in container
    $this->instance(Request::class, $this->prepareRequest($request));

    return [$request->getMethod(), '/'.trim($request->getPathInfo(), '/')];
}
```

So, I think it can be remove after version v5.7.6.

taylorotwell:
I don't have a problem leaving it for this point release at least.

PR #855, opened 2018-12-04
[5.7] Update Router.php

iocaste:
Opposite to laravel, lumen at the moment always appends namespace, this PR makes lumen router to check, if class starts with backslash (same as laravel), if it starts with backslash, namespace shouldn't be applied to class name.

### DIFF
--- a/src/Routing/Router.php
+++ b/src/Routing/Router.php
@@ -270,15 +270,14 @@ class Router
     /**
      * Prepend the namespace onto the use clause.
      *
-     * @param $class
-     * @param $namespace
-     *
+     * @param  string  $class
+     * @param  string $namespace
      * @return string
      */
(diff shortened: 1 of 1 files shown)


src/Routing/Router.php
```diff
@@ -260,13 +260,27 @@ protected function mergeGroupAttributes(array $action, array $attributes)
...
+            $action['uses'] = $this->prependGroupNamespace($action['uses'], $namespace);
         }
 
         return $action;
     }
 
+    /**
+     * Prepend the namespace onto the use clause.
+     *
+     * @param $class
```
GrahamCampbell: invalid syntax

iocaste:
@GrahamCampbell I've updated PR

taylorotwell:

Thanks!

PR #863, opened 2018-12-19
[5.7] Support app config for URL, with env var fallback

jmauerhan:
When the line setting the root URL was added, this forced use of the APP_URL env var, instead of using the same value that we can set in `config/app.php`, which can be any name for the environment variable. So if you had a different variable name such as BASE_URL (I suspect this is an artifact of an older version of lumen), and used that to set the `url` value in the app config file, the tests no longer worked, because the tests force you to use APP_URL. 

This change would default to whatever is set in app.php, with a fallback to the APP_URL value.

### DIFF
--- a/src/Testing/TestCase.php
+++ b/src/Testing/TestCase.php
@@ -55,7 +55,9 @@ abstract class TestCase extends BaseTestCase
 
         $this->app = $this->createApplication();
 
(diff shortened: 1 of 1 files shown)


jmauerhan:
@GrahamCampbell actually raised this issue when the code was added: https://github.com/laravel/lumen-framework/pull/825

PR #865, opened 2019-01-03
[5.8] Use similar method as Laravel UrlGenerator::formatParameters().

crynobone:
Signed-off-by: Mior Muhammad Zaki <crynobone@gmail.com>

### DIFF
--- a/src/Routing/UrlGenerator.php
+++ b/src/Routing/UrlGenerator.php
@@ -2,6 +2,7 @@
 
 namespace Laravel\Lumen\Routing;
 
+use Illuminate\Support\Arr;
 use Laravel\Lumen\Application;
 use Illuminate\Contracts\Routing\UrlRoutable;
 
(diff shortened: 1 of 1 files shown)


src/Routing/UrlGenerator.php
```diff
@@ -324,18 +345,11 @@ protected function formatParametersForUrl($parameters)
...
-        $parameters = is_array($parameters) ? $parameters : [$parameters];
-
-        foreach ($parameters as $key => $parameter) {
-            if ($parameter instanceof UrlRoutable) {
-                $parameters[$key] = $parameter->getRouteKey();
-            }
-        }
-
-        return $parameters;
+        return $this->formatParameters($parameters);
```
crynobone: Both `formatParametersForUrl()` and `replaceRoutableParametersForUrl()` is only used internally, but I'm not sure how Lumen deprecation structure works on I just mark it for deprecated at the moment.

driesvints: Imo can be removed since it's for a new release.

crynobone: Done

PR #883, opened 2019-02-27
[5.8] ConsoleServiceProvider implement DeferrableProvider contract

LarsGrevelink:
In Laravel 5.8 the `$defer` boolean property on the provider is deprecated. Marking the provider as deferred now works via the `Illuminate\Contracts\Support\DeferrableProvider`contract.

https://laravel.com/docs/master/upgrade#deferred-service-providers


driesvints:
Lumen doesn't even defers providers so this change wouldn't matter. I think it's best that we just remove the defer property on the `ConsoleServiceProvider`. 

https://github.com/laravel/lumen-framework/blob/fc78ca0b0311b4a596ba72851c1d2519e54b5439/src/Application.php#L207-L210

LarsGrevelink:
Was aware but agreed. If a future implementation should require this provider to be deferred it would make more sense to add it when Lumen provides support for it. Will revert the additions.

PR #893, opened 2019-03-12
[5.8] Removed should receive 'fire' from expectsEvents and withoutEvents

DaanGeurts:
When mocking events it is still checking if the fire method is triggered but this method has been deprecated since Laravel 5.4 and removed since 5.8 so the test passes while the code will fail.


driesvints:
Thanks!

PR #900, opened 2019-03-22
[5.8] Removed annoying semi column from json method.

stefanlazarevic:
Removed the unnecessary semi column from the documentation. VSCode intelephense complains about it and PHP7 return type feature can't be used because putting semi-column in there is not syntactically correct.

P.S I already removed it locally, and it works like a charm.

Screenshot of mentioned issue: https://ibb.co/zn61dsV


driesvints:
Thanks!

PR #903, opened 2019-04-02
[5.8] Fix passing an empty string to the predis client

arunfung:
When specifying the index, make sure to put the 0 in quotes, otherwise it just gets evaluated as false which ends up passing an empty string to the predis client.
REDIS_DATABASE = 0 gets turned into "database":"" and gives ERR invalid DB index
REDIS_DATABASE = "0" results in "database":"0" and works

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -115,14 +115,14 @@ return [
             'host' => env('REDIS_HOST', '127.0.0.1'),
             'password' => env('REDIS_PASSWORD', null),
             'port' => env('REDIS_PORT', 6379),
-            'database' => env('REDIS_DB', 0),
+            'database' => env('REDIS_DB', '0'),
         ],
 
         'cache' => [
             'host' => env('REDIS_HOST', '127.0.0.1'),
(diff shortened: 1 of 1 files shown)


driesvints:
This has been like this for a while, as well as on the laravel skeleton. Are you sure this is the correct way to go? https://github.com/laravel/laravel/blob/master/config/database.php

arunfung:
> This has been like this for a while, as well as on the laravel skeleton. Are you sure this is the correct way to go? https://github.com/laravel/laravel/blob/master/config/database.php



> This has been like this for a while, as well as on the laravel skeleton. Are you sure this is the correct way to go? https://github.com/laravel/laravel/blob/master/config/database.php

the problem has been replicated and been verified in the driver "predis"

driesvints:
I see you've changed the base branch. I'm going to have to ask you to resubmit because there's a bug in Github atm with changing the base branch which can cause unwanted commits to be merged.

arunfung:
I'm sorry. After verification again, I found that this problem does not exist in 5.8

PR #906, opened 2019-04-08
[5.8] Allow filesystem factory to be dependency injectable

Kieran-Brown:
Lumen adds the filesystem factory as an available binding but as there is no alias the resolution fails.
https://github.com/laravel/lumen-framework/blob/32fa445283c182960891f4f21355376cf0ea3b31/src/Application.php#L937-L938

### DIFF
--- a/tests/FullApplicationTest.php
+++ b/tests/FullApplicationTest.php
@@ -610,6 +610,15 @@ class FullApplicationTest extends TestCase
         $this->assertSame('1234', $response->getContent());
     }
 
+    public function testCanResolveFilesystemFactoryFromContract()
+    {
+        $app = new Application();
+
+        $filesystem = $app['Illuminate\Contracts\Filesystem\Factory'];
+
+        $this->assertInstanceOf('Illuminate\Contracts\Filesystem\Factory', $filesystem);
+    }
+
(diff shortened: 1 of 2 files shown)


driesvints:
Please provide an explanation why this is needed and add a test for this.

tests/FullApplicationTest.php
```diff
@@ -610,6 +610,15 @@ public function testRequestUser()
         $this->assertSame('1234', $response->getContent());
     }
 
+    public function testCanResolveFilesystemFactoryFromContract()
+    {
+        $app = new Application();
+
+        $filesystem = $app['Illuminate\Contracts\Filesystem\Factory'];
+
+        $this->assertInstanceOf('Illuminate\Contracts\Filesystem\Factory', $filesystem);
```
driesvints: Please use the `::class` notation for this and the above reference.

Kieran-Brown: @driesvints most of the tests here don't use the `::class` notation. I've updated this one but it might be worth updating the rest of the tests too.

driesvints: @Kieran-Brown yeah, I'm aware. Just want to prevent any more from getting in. Thanks!

PR #909, opened 2019-04-16
[5.8] Fix setting of APP_ENV variable in tests

sml-joyo:
Environment variables are loaded into the `$_ENV` superglobal, not the `putenv` store. As is, this does not modify or update the _correct_ environment variable resulting in `app()->environment()` and other methods returning the "wrong" value.

### DIFF
--- a/src/Testing/TestCase.php
+++ b/src/Testing/TestCase.php
@@ -49,7 +49,7 @@ abstract class TestCase extends BaseTestCase
      */
     protected function refreshApplication()
     {
-        putenv('APP_ENV=testing');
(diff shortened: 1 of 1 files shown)


taylorotwell:
Seems like a dangerous thing to change on a patch release.

sml-joyo:
I can target this against 5.9 if preferred. I did deliberate whether it would be considered breaking, but given it only affects tests and the default is `testing` I figured it wasn't significant.

taylorotwell:
image

Returns `testing`...

PR #937, opened 2019-07-11
[5.8] Remove useless catch exception.

sanchescom:
HttpResponseException was handled before \Exceptions\Handler. It will never go to render() method.


taylorotwell:
No plans to change anything related to this on a patch release.

PR #952, opened 2019-07-31
[5.8] Match Laravels Exception Handler when JSON is expected

koenwoortman:
When a JSON response is expected, return an instance of JsonResponse in the Exception Handler. This matches the behavior of Laravel.

For more information see: https://github.com/laravel/lumen-framework/pull/945


taylorotwell:
This would need to go to master branch. Thanks.

PR #961, opened 2019-09-04
Target 6.0

Juhlinus:
Updated `composer.json` and `src/Application.php` to target version 6.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: php
 
 php:
-  - 7.1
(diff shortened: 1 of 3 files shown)


Juhlinus:
I just realized that I shouldn't be merging this into the 5.8 branch, should I close this PR and target the master branch instead?

GrahamCampbell:
We already have a v6 branch at https://github.com/laravel/lumen-framework/blob/master/composer.json.

PR #978, opened 2019-09-24
[6.x] Use redis cache connection by default

gmsantos:
Update default redis cache connection to use this specific connection.

The cache connection is defined in databases config (https://github.com/laravel/lumen-framework/blob/6.x/config/database.php#L121)

### DIFF
--- a/config/cache.php
+++ b/config/cache.php
@@ -72,7 +72,7 @@ return [
 
         'redis' => [
             'driver' => 'redis',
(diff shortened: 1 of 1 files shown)


taylorotwell:
I think this would be a breaking change?

gmsantos:
I think that in the worst scenario a cache hit will be missed and a new key will be created on the new database @taylorotwell

PR #985, opened 2019-10-03
[6.x] Allow multiple optional parameters in named routes

### LINKED ISSUE #982: route() returns invalid URL for named route with two optional parameters
- Lumen Version: 6.0.2
- Laravel Version: 6.0
- PHP Version: 7.3.10

### Description:

`route()` returns invalid URL for named route with two optional params

### Steps To Reproduce:

I have two routes. First is main route for home page (url: `/`), second is for test and have two optional params - `id` and `name`. I want to make redirect from home page to named route "test", but `route()` helper returns invalid URL

This is my code:
```php
/**
 * @var Laravel\Lumen\Routing\Router $router
 */

$router->get('/', function () {
    return route('test');
    // return redirect()->route('test');
});

$router->get('/test[/{id}[/{name}]]', [
    'as' => 'test', function ($id = 0, $name = '') {
        if ($name) {
            return "Name: $name (ID: $id)";
        }
        if ($id) {
            return "ID: $id";
        }
        return 'Need an id or name!';
    }
]);
```

This is what returns `route()` helper for route test: `https://example.com/test[/{id}[/{name}]]`

sebastianthiel:
Closes #982.

**Before**
using `UrlGenerator->route()` on a Url like `/test[/{foo}[/{bar}]]`  generates a URL like `https://example.com/test[/{foo}[/{bar}]]`.

**After**
using `UrlGenerator->route()` on a Url like `/test[/{foo}[/{bar}]]`  generates a URL like `https://example.com/test`.

Parameters will also be handled in the correct order.
E.g.
`/test[/{foo}[/{bar}]]` with `foo = 1` returns `https://example.com/test/1`
`/test[/{foo}[/{bar}]]` with `foo = 1 and bar = 2 ` returns `https://example.com/test/1/2`
`/test[/{foo}[/{bar}]]` with `bar = 2 ` returns `https://example.com/test?bar=2`

### DIFF
--- a/src/Routing/UrlGenerator.php
+++ b/src/Routing/UrlGenerator.php
@@ -235,7 +235,7 @@ class UrlGenerator
 
         $parameters = $this->formatParameters($parameters);
 
-        $uri = $this->replaceOptionalParts($uri,$parameters);
+        $uri = $this->replaceOptionalParts($uri, $parameters);
 
         $uri = $this->replaceRouteParameters($uri, $parameters);
 
@@ -249,7 +249,7 @@ class UrlGenerator
     }
 
     /**
-     * handle optional parts in a route
+     * handle optional parts in a route.
      *
      * @param string $uri
      * @param array  $parameters
@@ -266,7 +266,7 @@ class UrlGenerator
                     return '';
                 }
 
-                return (count($matches) === 3) ? $uri . $this->replaceOptionalParts($matches[2], $parameters) : $uri;
+                return (count($matches) === 3) ? $uri.$this->replaceOptionalParts($matches[2], $parameters) : $uri;
             },
             $uri);
     }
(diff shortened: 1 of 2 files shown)


driesvints:
You might want to add a thorough explanation to your pr so it's clear what this fixes. Don't just link to an issue.

sebastianthiel:
thanks for the hint, I updated the first comment, hope it is ok.

taylorotwell:

sebastianthiel:
Hi @taylorotwell, 

the documentation talks about optional parameter**s**, enclosed in `[...]`. It does not say if multiple levels are allowed, but either case more than one optional parameter are not nice to handle.


*E.g.*
url: `/foo[/{bar}/{baz}]`
since both parameter are optional, it should be also possible to have only one of them. But this url with bar set to 1 would result into `/foo/1/{baz}`.

so the next logical step would be to put the `baz` into a optional optional part like ``/foo[/{bar}[/{baz}]]``. But this will break the the router completely.


But on the other hand, this has not worked before, so this could also be considered as a new feature.

As you said before, you're the one who has to maintain the code. So the decision is up to you. Either way, I had some fun with regex :smiley:

PR #1025, opened 2020-01-25
[6.x] Fix directory separator

daarond:
use directory separator constant instead of hard-coded string. breaks windows paths.


GrahamCampbell:
Do you have an example where it actually "doesn't work"?

daarond:
you're right, I was a bit hasty adding this one when the real bug is in File::getTargetFile

PR #1026, opened 2020-01-29
[6.x] Support changing pg database timezone

beuss:
Since PostgresConnector supports it, allow to set DB timezone through
DB_TIMEZONE .env entry


taylorotwell:
You're free to add this to your own configuration file if needed.

beuss:
It is supported out of the box for mysql, why not for pg?

PR #1033, opened 2020-02-19
Mistake in \Laravel\Lumen\Exceptions\Handler::render() return type hint

HenkPoley:
It should match the return typehint in \Illuminate\Contracts\Debug\ExceptionHandler

This resolves `vimeo/psalm` always complaining about InvalidReturnStatement, LessSpecificImplementedReturnType, or ImplementedReturnTypeMismatch in your `app/Exceptions/Handler.php` no matter what you do.

Installed contract filename is  illuminate/contracts/Debug/ExceptionHandler.php, found here: https://github.com/laravel/framework/blob/6.x/src/Illuminate/Contracts/Debug/ExceptionHandler.php

### DIFF
--- a/src/Exceptions/Handler.php
+++ b/src/Exceptions/Handler.php
@@ -87,7 +87,7 @@ class Handler implements ExceptionHandler
      *
      * @param  \Illuminate\Http\Request  $request
      * @param  \Exception  $e
-     * @return \Illuminate\Http\Response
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Thanks for this. It looks like the entire file is a bit out of sync with the contract. I'll update that shortly.

PR #1038, opened 2020-02-23
[7.x] Use Illuminate\Testing\TestResponse

crynobone:
Signed-off-by: Mior Muhammad Zaki <crynobone@gmail.com>

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,6 +33,7 @@
         "illuminate/pipeline": "^7.0",
         "illuminate/queue": "^7.0",
         "illuminate/support": "^7.0",
+        "illuminate/testing": "^7.0",
         "illuminate/translation": "^7.0",
         "illuminate/validation": "^7.0",
         "illuminate/view": "^7.0",
@@ -48,7 +49,7 @@
     },
     "require-dev": {
         "mockery/mockery": "^1.0",
-        "phpunit/phpunit": "^8.0"
+        "phpunit/phpunit": "^8.4 || ^9.0"
     },
     "suggest": {
         "laravel/tinker": "Required to use the tinker console command (^2.0).",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -48,7 +49,7 @@
     },
     "require-dev": {
         "mockery/mockery": "^1.0",
-        "phpunit/phpunit": "^8.0"
+        "phpunit/phpunit": "^8.4 || ^9.0"
```
GrahamCampbell: Please use the single bar without spaces, syntax.

composer.json
```diff
@@ -33,6 +33,7 @@
         "illuminate/pipeline": "^7.0",
         "illuminate/queue": "^7.0",
         "illuminate/support": "^7.0",
+        "illuminate/testing": "^7.0",
```
GrahamCampbell: Please move this to `require-dev` and make it a "suggested" dependency.

crynobone: > Please move this to require-dev and make it a "suggested" dependency.

If you want to make a HTTP tests on your Lumen app, then you need to manually add the dependency, which is different to Laravel where this would be available by default.

GrahamCampbell: Yes, that is correct. Lumen is intentionally light out of the box. Lots of stuff doesn't work unless you opt in to it.

GrahamCampbell: We should include this dependency in the example `laravel/lumen` app's require-dev too.

GrahamCampbell: And in the upgrading guide.

driesvints:
Isn't it best that we add it to "require" instead?

taylorotwell:
@crynobone yeah, seems like we would want to require this? Otherwise it won't be available in people's applications by default?

GrahamCampbell:
Now that phpunit is no longer required by illuminate/testing, I am not against moving this back into the framework's own require block. We could go with either solution tbh.

PR #1041, opened 2020-03-08
Allow use of static functions for routing methods

rentalhost:
- [x] It is a bug, considering that Laravel works with same code;
 - [x] Allow use of static functions for routing methods;
 - [x] Tests updated and passing; 
 - [x] Respecting code style;

Currently, routing methods allow only an anonymous non-static functions, like:

```php
$app->router->get('/', function () {
    return response('Hello World');
});
```

When we tried to use a static function, once that we do not need `$this` access in some cases, we will get an exception:

 > **ErrorException (E_WARNING)**
    Cannot bind an instance to a static closure

But sometimes we really do not need `$this` (as example of the test units), so we could do it that way:

```php
$app->router->get('/', static function () {
    return response('Hello World');
});
```

To resolve that, I have modified the `RoutesRequests.php` file to `try {}` to bind first, but in case of fail, it will use the own value itself, statically.

I have updated a testing to use `static function` and it passes. Other tests still uses non-static functions and keep working too.

### DIFF
--- a/src/Concerns/RoutesRequests.php
+++ b/src/Concerns/RoutesRequests.php
@@ -280,7 +280,11 @@ trait RoutesRequests
 
         foreach ($action as $value) {
             if ($value instanceof Closure) {
-                $closure = $value->bindTo(new RoutingClosure);
+                try {
+                    $closure = $value->bindTo(new RoutingClosure);
+                } catch (\ErrorException $exception) {
+                    $closure = $value;
+                }
                 break;
             }
         }
--- a/tests/FullApplicationTest.php
+++ b/tests/FullApplicationTest.php
@@ -17,7 +17,7 @@ class FullApplicationTest extends TestCase
     {
         $app = new Application;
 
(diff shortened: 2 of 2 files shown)


taylorotwell:

dunhamjared:
@rentalhost Did you experience any speed improvements?

rentalhost:
@dunhamjared actually I have done some benchmarks and I could see that when using `static function` it is 6x slower than just `function`, due the `->bindTo()` throws an exception that should be handled by the method that I have implemented.

I have did other tests with `ReflectionFunction` and it is 2x slower, but will affect both `function` and `static function`, but the slowdown is more reasonable, except because it will not add any real advantage (we could simply use `function`, even if `static function` will works fine).

Anyway, the real performance affected is imperceptible. For instance, it will use `0.000001 s` to handle the `function` and `0.000006 s` to handle `static function`. And it happen a single time by request.

PR #1049, opened 2020-03-16
add ramsey/uuid dependency,because of the illuminate/queue  rely on S…

zeevin:
add ramsey/uuid dependency,because of the illuminate/queue  rely on Str::uuid,but  in illuminate/support package it was just an suggest package.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,6 +40,7 @@
         "illuminate/log": "^7.0",
         "dragonmantank/cron-expression": "^2.0",
         "nikic/fast-route": "^1.3",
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
This PR was reverted. The correct fix is to make the uuid package a proper requirement in the queue package, since it is not so optional anymore.

zeevin:
yes, i wanted to do that before ,but lluminate/queue  repo is   [READ ONLY].

GrahamCampbell:
Because the repo is generated from laravel/framework.

PR #1060, opened 2020-03-28
[7.x] Fixed receive json

### LINKED ISSUE #1059: Method receiveJson fails when verifying response content
- Lumen Version: 7.0.2
- PHP Version: 7.4.4

### Description:

We use the `receiveJson` method to ensure that our API returns json content.

### Steps To Reproduce:

1. Write a test as following:

```php
/**
 * @test
 */
public function it_should_return_json_content(): void
{
     $this->get(route('route.name', ['per_page' => 10]))
          ->seeStatusCode(Response::HTTP_OK)
          ->receiveJson();
}
```
2. Run the test
3. You should see the following error:
```shell
There was 1 error:

1) Tests\Unit\SampleTest::it_should_return_json_content
TypeError: Argument 1 passed to Laravel\Lumen\Testing\TestCase::seeJsonContains() must be of the type array, null given, called in .../vendor/laravel/lumen-framework/src/Testing/Concerns/MakesHttpRequests.php on line 254
```

Pr3d4dor:
Closes #1059

I don't know how to write a test for this. If someone could help me, I will be happy to add a test.

### DIFF
--- a/src/Testing/Concerns/MakesHttpRequests.php
+++ b/src/Testing/Concerns/MakesHttpRequests.php
@@ -249,6 +249,8 @@ trait MakesHttpRequests
                     "JSON was not returned from [{$this->currentUri}]."
                 );
             }
+
+            return $this->seeJsonContains($decodedResponse, $negate);
         }
 
         return $this->seeJsonContains($data, $negate);


driesvints:
Closing this one over #1061 because it has tests. Thanks!

PR #1062, opened 2020-03-30
[7.x] Fixes #1055

### LINKED ISSUE #1055: php artisan --env= doesn't affect execution
- Lumen Version: 6.3.3/7.0.2
- Laravel Version: ^6.0 / ^7.0
- PHP Version: 7.3.9
- OS: Windows 10

### Description:
the command artisan <some command> --env=testing doesn't affect the environment of the current call.

### Steps To Reproduce:
Create an empty lumen application in windows. Configure a sqlite connection in .env:
```
DB_CONNECTION=sqlite
DB_DATABASE=C:\test.sqlite
```

1. `php artisan migrate` => no errors.
2. copy the `.env` to `.env.testing`. 
3. Change the Path to an invalid value in the `.env`
4. `php artisan migrate --env=testing` =>path from the .env.testing should be taken

What happens instead?
the value from the .env is taken. in laravel this works without problems.

bytestream:
### Description

Fixes:
* use of `php artisan <command> --env <env>` not using the `--env` param.
* use of `.env.<environment>` files, previously it would only ever load `.env`

This is the most BC way I could implement a fix. Laravel loads dotenv after creating an Application instance while Lumen is doing the complete opposite.

### Tests

laravel/framework has no tests for these functions.

### DIFF
--- a/src/Bootstrap/LoadEnvironmentVariables.php
+++ b/src/Bootstrap/LoadEnvironmentVariables.php
@@ -5,6 +5,8 @@ namespace Laravel\Lumen\Bootstrap;
 use Dotenv\Dotenv;
 use Dotenv\Exception\InvalidFileException;
 use Illuminate\Support\Env;
+use Laravel\Lumen\Application;
+use Symfony\Component\Console\Input\ArgvInput;
 use Symfony\Component\Console\Output\ConsoleOutput;
 
 class LoadEnvironmentVariables
@@ -27,10 +29,10 @@ class LoadEnvironmentVariables
      * Create a new loads environment variables instance.
      *
      * @param  string  $path
-     * @param  string|null  $name
+     * @param  string  $name
      * @return void
      */
-    public function __construct($path, $name = null)
+    public function __construct($path, $name = '.env')
     {
         $this->filePath = $path;
         $this->fileName = $name;
@@ -45,6 +47,8 @@ class LoadEnvironmentVariables
      */
     public function bootstrap()
     {
+        $this->checkForSpecificEnvironmentFile();
+
         try {
             $this->createDotenv()->safeLoad();
         } catch (InvalidFileException $e) {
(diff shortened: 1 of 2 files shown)


bytestream:
Not sure - I only just started using it. Most likely answer is no - https://github.com/laravel/lumen/blob/5.7/bootstrap/app.php#L5

src/Bootstrap/LoadEnvironmentVariables.php
```diff
@@ -27,10 +29,10 @@ class LoadEnvironmentVariables
      * Create a new loads environment variables instance.
      *
      * @param  string  $path
-     * @param  string|null  $name
+     * @param  string  $name
      * @return void
      */
-    public function __construct($path, $name = null)
+    public function __construct($path, $name = '.env')
```
GrahamCampbell: This is a breaking change.

GrahamCampbell:
> Not sure - I only just started using it.

I would probably recommend new apps start off using the full framework. It is just as fast, if not faster if you have a lot of routes.

bytestream:
> I would probably recommend new apps start off using the full framework.

What's the point of Lumen then? heh. I use the full framework elsewhere. This project is just a very simple API.

---

#1055 is not such an issue for me. I only briefly ran into it, noticed it tagged as a bug and offered a fix. I don't mind manually overwriting `.env` with `.env.<name>`

It should be documented though if it's not supported, particularly the `--env <name>` on console commands.

src/Bootstrap/LoadEnvironmentVariables.php
bytestream: How so? `null` becomes `.env` upstream.

https://github.com/vlucas/phpdotenv/blob/master/src/Store/StoreBuilder.php#L98

kanidjar:
> I would probably recommend new apps start off using the full framework. It is just as fast, if not faster if you have a lot of routes.

This is quite disturbing to read something like this from a team member. If you really think that Lumen has now become useless, you should agree together to stop its development.

GrahamCampbell:
> This is quite disturbing to read something like this from a team member. If you really think that Lumen has now become useless, you should agree together to stop its development.

That's not what I said. I just wanted to express the fact that for 99.9% of use cases, the full framework is probably the best place to start. A lot of effort has been made over the course of Laravel 5 and 6 to make it even faster, in particular with respect to routing in Laravel 7. These routing improvements have not come to Lumen, so this should certainly be a consideration for new projects.

GrahamCampbell:
I myself do use Lumen, so. ;)

taylorotwell:
Going to hold off on this for now.

DBX12:
Sad to see this closed while the console command `php artisan migrate --help` still advertises the `--env` switch.

PR #1064, opened 2020-04-11
[8.x] Version 8

GrahamCampbell:
The version constant in the application class has already been updated to version 8, but the branch alias says v7. Not good, since people asking for v7 dev versions will get something which calls itself v8, but is actually v7. Either, we should reset master to be the same as the 7.x branch, or we should merge this PR (or something like it).


driesvints:
Thanks

PR #1081, opened 2020-05-27
[7.x] Replaced array_walk with foreach

mrjeanp:
foreach performs 33% faster than array_walk

https://medium.com/@steveclifton_12558/array-walk-performance-test-b8cfb11511a8

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -225,9 +225,9 @@ class Application extends Container
             return;
         }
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Did you notice an actual difference in a real app?

mrjeanp:
Not at this moment, but it won't break anything. 
I'm developing a project and i'm trying to optimize as much as possible

PR #1094, opened 2020-06-20
[7.x] Remove redundant return from void method

antonkomarev:
Follow up #1092 

Removing return because `register` method returns nothing.


driesvints:
Thanks

PR #1098, opened 2020-06-22
[7.x] Reorder register and boot methods

antonkomarev:
This PR reordering `register` and `boot` methods in EventServiceProvider.

There are 2 reasons behind it:
1. `register` method is executing before the `boot` method
2. it's similar to the Laravel code base


driesvints:
Thanks!

PR #1103, opened 2020-08-12
[7.x] Update Handler.php

ibpavlov:
Fix exception handler to report general exceptions that cannot be converted to a string automatically.


GrahamCampbell:
Thanks, this also matches what the laravel/framework repo does.

PR #1114, opened 2020-09-18
[8.x] Register BatchesTableCommand in ConsoleServiceProvider

kanidjar:
Register BatchesTableCommand so it is available when using artisan.


taylorotwell:
@GrahamCampbell yes, we will support it.

PR #1124, opened 2020-10-31
[6.x] Add support for php 8

chris-doehring:
This PR will add php 8 support for Lumen 6.x. I used the same version constraints as in [Laravel 6.x](https://github.com/laravel/framework/blob/6.x/composer.json) for the relevant packages.


GrahamCampbell:
Lumen 6 is not an LTS release. Only laravel/framework was.

PR #1133, opened 2020-11-19
[8.x] Default Redis cache connection is wrong

MisterCoder:
Cache clearing when using Redis would also clear the job queue as the default connection is used by both.

See https://github.com/laravel/laravel/blob/8.x/config/cache.php#L77


driesvints:
This is a breaking change in a stable release and should probably target master.

driesvints:
Think it's best that you try a fresh PR.

MisterCoder:
Created https://github.com/laravel/lumen-framework/pull/1134

PR #1145, opened 2021-01-07
[9.x] Enable support of loading multiple dotenv files

nabihnadhif:
We use Lumen to develop many microservices and deploy them over Microsoft Azure DevOps platform.
Some environment variables are common between all microservices. So, for maintenance purpose, it will be more easier to mutualise them on a common dotenv file (.env.shared for example) and to allow loading app specific dotenv file with the common one.
This PR should allow that. I Added tests to ensure compatibility with existing apps.


taylorotwell:

PR #1146, opened 2021-01-15
Update RegistersExceptionHandlers.php

MohamadSH:
Adding an integer in order to pass to HttpException for using in making appropriate API responses.
Imagine a condition which you want to abort a request with 400 http status code and also set an integer code for interpreting the response

### DIFF
--- a/src/Concerns/RegistersExceptionHandlers.php
+++ b/src/Concerns/RegistersExceptionHandlers.php
@@ -16,20 +16,21 @@ trait RegistersExceptionHandlers
     /**
      * Throw an HttpException with the given data.
      *
-     * @param  int  $code
+     * @param  int  $statusCode
      * @param  string  $message
      * @param  array  $headers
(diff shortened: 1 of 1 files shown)


src/Concerns/RegistersExceptionHandlers.php
```diff
@@ -16,20 +16,21 @@ trait RegistersExceptionHandlers
...
+     * @param  int  $statusCode
      * @param  string  $message
      * @param  array  $headers
+     * @param  int  $code
      * @return void
      *
      * @throws \Symfony\Component\HttpKernel\Exception\HttpException
      */
-    public function abort($code, $message = '', array $headers = [])
+    public function abort($statusCode, $message = '', array $headers = [] , int $code = 0)
```
driesvints, line 27: Renaming parameter names is a breaking change for PHP 8 as well as is the extra parameter addition. Please send this to master.

PR #1148, opened 2021-01-21
[8.x] Adding missing headers when exception is of type HttpException.

nxuls:
Adding functionality that was present in 6.x to add headers to the response object in the exception handler, when the exception is an `HttpException`. This was a breaking issue for an application I develop for when I attempted to upgrade the Lumen version. It is not documented in the upgrade guide that this is new expected behavior.

### DIFF
--- a/src/Exceptions/Handler.php
+++ b/src/Exceptions/Handler.php
@@ -165,7 +165,8 @@ class Handler implements ExceptionHandler
     {
         $response = new Response(
             $this->renderExceptionWithSymfony($e, config('app.debug', false)),
-            $this->isHttpException($e) ? $e->getStatusCode() : 500
+            $this->isHttpException($e) ? $e->getStatusCode() : 500,
+            $this->isHttpException($e) ? $e->getHeaders() : []
         );
 
         $response->exception = $e;


taylorotwell:
@GrahamCampbell @driesvints any idea why this would have been removed?

GrahamCampbell:
Looks like it was either done for consistency with laravel/framework, or by mistake: https://github.com/laravel/lumen-framework/commit/c354d403555a4d311bb3f910e3eddebee773516c.

GrahamCampbell:
We should check if there are similar errors in the framework codebase, and indeed, in other places in lumen.

driesvints:
Seems like we always pass the headers in laravel/framework:

https://github.com/laravel/framework/blob/6e6d699b05c2fe9447e26adaf26e154c8063a34d/src/Illuminate/Foundation/Exceptions/Handler.php#L477
https://github.com/laravel/framework/blob/6e6d699b05c2fe9447e26adaf26e154c8063a34d/src/Illuminate/Foundation/Exceptions/Handler.php#L595-L599

So seems reasonable to keep doing it here as well.

driesvints (approved):

GrahamCampbell (approved):

PR #1176, opened 2021-07-02
Make default timezone to look in .env APP_TIMEZONE

jmcnualan:


driesvints:
Please see many previous attempts. This isn't a setting that changes per environment.

PR #1177, opened 2021-07-12
fix empty response exception

sashsvamir:
fix issue #1178 – empty response exception when response method of exception return void result (eg `Illuminate\View\ViewException`)


driesvints:
If there's a view exception then there's a bug in your code. Please try a support channel.

sashsvamir:
@driesvints Why lumen is handles another exceptions (`ModelNotFoundException` and etc), and show for them debug stack trace, but not `ViewException`? If lumen has support of blade templates, then it would be good to gives to developer trace info on error happens.

PR #1187, opened 2021-09-02
[8.x] Fixed DocBlocks

dunhamjared:
- Replaced `@inheritDoc` -- missing parent member with doc comments 
- Updated docblock for `gatherMiddlewareClassNames` -- the `$middleware` parameter can handle arrays and the [`callLumenControllerWithMiddleware`](https://github.com/laravel/lumen-framework/blob/8.x/src/Concerns/RoutesRequests.php#L366) function passes an array

### DIFF
--- a/src/Concerns/RoutesRequests.php
+++ b/src/Concerns/RoutesRequests.php
@@ -85,7 +85,10 @@ trait RoutesRequests
     }
 
     /**
-     * {@inheritdoc}
+     * Dispatch request and return response
+     *
+     * @param  SymfonyRequest $request
+     * @return Response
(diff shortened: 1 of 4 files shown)


taylorotwell:
Please always use fully qualified class names in doc blocks.

driesvints:
Mark this as ready when you've converted to FQN, thanks.

dunhamjared:
Ok, I sifted through the `src` directory and found a few more doc blocks that needed fully qualified class names. 

Also discovered two more `@inheritdoc` that were missing parent member with doc comments.

PR #1201, opened 2021-11-07
Update helpers.php

vook:
Maybe if we use generics to guarantee autocomplete on certain IDEs

### DIFF
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -29,9 +29,10 @@ if (! function_exists('app')) {
     /**
      * Get the available container instance.
      *
-     * @param  string|null  $make
+     * @template T
+     * @param  class-string<T>  $make
      * @param  array  $parameters
-     * @return mixed|\Laravel\Lumen\Application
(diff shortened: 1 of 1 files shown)


src/helpers.php
```diff
@@ -29,9 +29,10 @@ function abort($code, $message = '', array $headers = [])
     /**
      * Get the available container instance.
      *
-     * @param  string|null  $make
+     * @template T
+     * @param  class-string<T>  $make
```
GrahamCampbell, line 33: this doc isn;t accurate i'm afraid. you can also put non-class-strings in here, like `hash` or `db.connection` (disclaimer: probs something like that :trollface:)

GrahamCampbell (changes requested):

src/helpers.php
GrahamCampbell: I think instead we should leave this to larastan and similar tools to implement cc @nunomaduro

PR #1205, opened 2021-12-01
[8.x] Fix PHP 8.1 deprecation issue

michalbundyra:
```
laravel.WARNING: trim(): Passing null to parameter #1 ($string) of type string is deprecated in /srv/api/vendor/laravel/lumen-framework/src/Routing/Router.php on line 151
```


chrismou:
:+1:

PR #1206, opened 2021-12-08
[8.x] Add Illuminate\Config\Repository::class as alias for config

mtracz:
In this PR, I added `Illuminate\Config\Repository::class` class as alias for `config` to the list of registered container aliases.

---
I discovered that `app()->make(Illuminate\Config\Repository::class);` does not return proper instance of config class (empty config values), actually found in PassportServiceProvider.
https://github.com/laravel/passport/blob/10.x/src/PassportServiceProvider.php#L268

LaravelPassport uses `Illuminate\Config\Repository::class` class for getting config values, so in this case if we are using Lumen, config is empty.
https://github.com/laravel/passport/blob/10.x/src/PassportServiceProvider.php#L8

I think we can do the same as Laravel do:
https://github.com/laravel/framework/blob/8.x/src/Illuminate/Foundation/Application.php#L1323

so
`app()->make(Illuminate\Config\Repository::class);`
and
`app()->make(Illuminate\Contracts\Config\Repository::class);`
will return the same config instance.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -1015,6 +1015,7 @@ class Application extends Container
             \Illuminate\Contracts\Cache\Factory::class => 'cache',
             \Illuminate\Contracts\Cache\Repository::class => 'cache.store',
             \Illuminate\Contracts\Config\Repository::class => 'config',
+            \Illuminate\Config\Repository::class => 'config',
             \Illuminate\Container\Container::class => 'app',
             \Illuminate\Contracts\Container\Container::class => 'app',
             \Illuminate\Database\ConnectionResolverInterface::class => 'db',


GrahamCampbell:
I'm actually kinda :-1: here. Why are you unable to use the contract?

mtracz:
Of course I can, but some packages as LaravelPassport in this case is not using contract for config in ServiceProvider, so it won't work correctly.
Also we can workaround and add `$app->alias("config", Illuminate\Config\Repository::class);` by yourself.

But why we won't to allow use `Illuminate\Config\Repository::class` and `Illuminate\Contracts\Config\Repository::class` by default as Laravel do?
Maybe this should be fixed in LaravelPassport and not here, I'm not sure.

GrahamCampbell:
Yep, I think we should fix passport here.

driesvints:
Why? Passport only works with Laravel so I'm not sure we should change anything there?

mtracz:
I thought, it can be changed, cause it won't bring any side effects. Also it will allow to use contract and non contract class for retrieving config values. Additionally it will solve problems such like in Passport.

The question was, why we won't to allow use Illuminate\Config\Repository::class and Illuminate\Contracts\Config\Repository::class by default as Laravel do?

Anyway, if you think that it is not worth to change it, then close this PR, and I will create PR in Passport.

taylorotwell:
It is aliased with both in the full Laravel framework so don't see a big reason not to do the same here.

PR #1210, opened 2021-12-28
[8.x] Add missing throws to doc blocks

dunhamjared:
Added missing throws to doc blocks.

Used `\Symfony\Component\HttpKernel\Exception\HttpException` for exceptions that extend `HttpsException` based on https://github.com/dunhamjared/lumen-framework/blob/8.x/src/Concerns/RegistersExceptionHandlers.php#L18-L35

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -921,8 +921,6 @@ class Application extends Container
      * Get the application namespace.
      *
      * @return string
-     *
-     * @throws \RuntimeException
      */
     public function getNamespace()
     {
--- a/src/Application.php
+++ b/src/Application.php
@@ -921,6 +921,8 @@ class Application extends Container
      * Get the application namespace.
      *
      * @return string
+     *
+     * @throws \RuntimeException
      */
     public function getNamespace()
     {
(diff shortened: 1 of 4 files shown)


GrahamCampbell:
Runtime exceptions should not go in throw doc. All methods throw RuntimeException|Error by default.

dunhamjared:
@GrahamCampbell 

Found two extra instances of RuntimeException and removed them. 

https://github.com/laravel/lumen-framework/blob/8.x/src/Http/Request.php#L56 - from 3 years ago
https://github.com/laravel/lumen-framework/blob/8.x/src/Application.php#L925 - from 6 years ago

src/Application.php
```diff
@@ -921,8 +921,6 @@ public function prepareForConsoleCommand($aliases = true)
      * Get the application namespace.
      *
      * @return string
-     *
-     * @throws \RuntimeException
```
GrahamCampbell, line 925: Please leave this fix to the master branch (if it was not already made there). The additional throw I think are fine for 8.x.

GrahamCampbell (changes requested):

dunhamjared:
Hi @taylorotwell 

Is there a reason for closing this pull request?

PR #1214, opened 2022-02-03
[6.x] PHP 8 Support

MarkVaughn:
For migrating from LTS version 6 to LTS version 9 there is no PHP version overlap.
Thanks to the work done here https://github.com/laravel/framework/pull/33388 and so much of the guts being shared, it seems safe to allow PHP 8 on lumen 6.x
This will help us with our migrations to PHP 8 so we can upgrade to Lumen 9 once it becomes available.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        php: [7.2, 7.3, 7.4]
+        php: [7.2, 7.3, 7.4, 8.0]
 
     name: PHP ${{ matrix.php }}
 
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
         }
     ],
     "require": {
-        "php": "^7.2",
+        "php": "^7.2|^8.0",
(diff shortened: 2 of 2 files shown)


GrahamCampbell:
👎 Lumen 6 is not LTS.

GrahamCampbell:
Only the main Laravel framework gets LTS releases.

MarkVaughn:
all the components lumen uses are PHP8 compatible, the choice to be on lumen 6 rather than lumen 7 benefits from the laravel 6s LTS fixes. This small change should have no impact, and really just seems to be an oversight

GrahamCampbell:
It's not an oversight. Lumen 6 is already EOL.

GrahamCampbell:
The tests passing here is not a good indication that Lumen 6 will work properly on PHP 8, and the wide version constraint adds the further problem that people will try to install it on PHP 8.1 too, where it definitely doesn't work. I think we're best of making people force work around the version constraint if they really wanna install it on PHP 8, so it's super clear it's not supported.

taylorotwell:
This would likely need more thorough review of the stuff that is Lumen specific and not part of Illuminate.

PR #1220, opened 2022-02-18
Inject container and config into ExceptionHandler for better testability

bambamboole:
First of all, thanks for all your efforts here :)

This PR injects the Container interface and the ConfigRepository interface into the constructor of the `Handler`.

This change allows easier testability when this class is overridden in a lumen application.
The current instantiation of the Handler is done via the Container, so there is no need to adapt anything.
I have adapted the Tests where the Handler was hacked and have replaced it with the real implementation.

If needed I can add some unit tests to the `Handler`.


taylorotwell:

PR #1222, opened 2022-02-21
Add langPath() method to Application

### LINKED ISSUE #1221: Composer package breaks project when using recommended app()->langPath() method
- Lumen Version: 8.3.4
- Laravel Version: 8.83.1
- PHP Version: 8.1.2
- Database Driver & Version: %

### Description:
I did not yet upgrade my Laravel and Lumen projects to version 9, but I did prepare my Laravel and Lumen packages for it.  
The [upgrade guide](https://laravel.com/docs/9.x/upgrade#the-lang-directory) recommends to set the path for package language files with `app()->langPath()` instead of hard-coding it. I changed that and did now a simple `composer update` in my Lumen project (still version 8) and it broke, because its [Application class](https://github.com/laravel/lumen-framework/blob/1b05a57a038866d1db10631dc2e5e3f5491cdf09/src/Application.php) does not contain a `langPath()` method. I'm surprised version 9 of the framework doesn't either.

### Steps To Reproduce:

In your Lumen project, pull in a Laravel/Lumen package with the following service provider:

```php
class MyPackageServiceProvider extends ServiceProvider
{
    public function boot(): void
    {
        $this->publishes([__DIR__ . '/../lang' => $this->app->langPath() . "/vendor/my-package"]);
    }
}
```

halloei:
Fixes #1221.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -832,6 +832,17 @@ class Application extends Container
     }
 
     /**
+     * Get the path to the language files.
+     *
+     * @param  string  $path
+     * @return string
+     */
+    public function langPath($path = '')
+    {
+        return $this->getLanguagePath().($path != '' ? DIRECTORY_SEPARATOR.$path : '');
+    }
+
+    /**
      * Get the storage path for the application.
      *
      * @param  string|null  $path


driesvints:
This needs to be sent in to Lumen v9. Lumen v8 isn't supported anymore.

PR #1230, opened 2022-03-24
Set ContainerCommandLoader for lazy loading console commands

MattApril:
This is necessary for using native Laravel commands, such as _queue:prune-batches._


hhxsv5:
@taylorotwell Could you please release a new version for this?

driesvints:
@hhxsv5 this has already been fixed with v9.0.2

PR #1234, opened 2022-04-20
Update base_path()

ElfSundae:
Avoid double-checking `$path`.


taylorotwell:
If current code is working it's fine.

PR #1238, opened 2022-06-14
fix: correct makes http request trait methods

santiagoibarzabal:
Updating the MakesHttpRequest trait so that many of its http helper methods return \Illuminate\Testing\TestResponse rather than the MakesHttpTrait class. Most of the methods are already executing $this->call() but return the complete trait rather than the TestResponse class. This works correctly in Laravel but not in Lumen.


taylorotwell:

Thanks!

santiagoibarzabal:
Will re-submit soon with a deeper explanation and adding tests. Thank you!

Issue #1240, opened 2022-07-20
Failing database migrations


kiwigod:
- Lumen Version: 9.0.3
- Laravel Version: #.#.#
- PHP Version: 8.1.8
- Database Driver & Version: sqlite

### Description:
All migration commands result in the following exception
```
developer@ed445aa882f3:/var/www$ php artisan -vvv migrate

In Component.php line 89:
                                                                           
  [Error]                                                                  
  Call to undefined function Illuminate\Console\View\Components\resolve()  
                                                                           

Exception trace:
  at /var/www/vendor/illuminate/console/View/Components/Component.php:89
 Illuminate\Console\View\Components\Component->mutate() at /var/www/vendor/illuminate/console/View/Components/Line.php:44
 Illuminate\Console\View\Components\Line->render() at /var/www/vendor/illuminate/console/View/Components/Info.php:18
 Illuminate\Console\View\Components\Info->render() at /var/www/vendor/illuminate/console/View/Components/Factory.php:56
 Illuminate\Console\View\Components\Factory->__call() at /var/www/vendor/illuminate/database/Console/Migrations/MigrateCommand.php:116
 Illuminate\Database\Console\Migrations\MigrateCommand->prepareDatabase() at /var/www/vendor/illuminate/database/Console/Migrations/MigrateCommand.php:79
 Illuminate\Database\Console\Migrations\MigrateCommand->Illuminate\Database\Console\Migrations\{closure}() at /var/www/vendor/illuminate/database/Migrations/Migrator.php:607
 Illuminate\Database\Migrations\Migrator->usingConnection() at /var/www/vendor/illuminate/database/Console/Migrations/MigrateCommand.php:103
 Illuminate\Database\Console\Migrations\MigrateCommand->handle() at /var/www/vendor/illuminate/container/BoundMethod.php:36
 Illuminate\Container\BoundMethod::Illuminate\Container\{closure}() at /var/www/vendor/illuminate/container/Util.php:41
 Illuminate\Container\Util::unwrapIfClosure() at /var/www/vendor/illuminate/container/BoundMethod.php:93
 Illuminate\Container\BoundMethod::callBoundMethod() at /var/www/vendor/illuminate/container/BoundMethod.php:37
 Illuminate\Container\BoundMethod::call() at /var/www/vendor/illuminate/container/Container.php:651
 Illuminate\Container\Container->call() at /var/www/vendor/illuminate/console/Command.php:139
[...]

PR #1241, opened 2022-07-20
Add resolve helper necessary for running migrations

kiwigod:
Currently it's not possible to run any sort of migrations due to the missing helper function. This pull request adds the helper function which can also be found in `laravel/framework` package under path `src/Illuminate/Foundation/helpers.php`

For reference see the specific function below:
https://github.com/laravel/framework/blob/eee833e46de5301aa533ce1b7f5372d375b00fd8/src/Illuminate/Foundation/helpers.php#L701

### DIFF
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -433,3 +433,17 @@ if (! function_exists('view')) {
         return $factory->make($view, $data, $mergeData);
     }
 }
+
+if (! function_exists('resolve')) {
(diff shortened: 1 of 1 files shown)


kiwigod:
For more context see issue https://github.com/laravel/lumen-framework/issues/1240
Since a pull request is created I'll close the issue now

PR #1246, opened 2022-08-19
Add optional $path parameter to app path() method

ElfSundae:


driesvints:
This is a breaking change to the method signature.

Issue #1247, opened 2022-09-08
excludeIf validation rule does not work as expected when using $this->validate from controller


MattApril:
- Lumen Version: 9.0.2
- Laravel Version: 9.17.0
- PHP Version: 8.1.6

### Description:
According to Laravel docs: 
> The field under validation will be excluded from the request data returned by the validate and validated methods.

In Lumen, this is not the case, and there is nothing stating so in the docs. It does respect the rule when applying other validation rules (a non-numeric value could but submitted in the example below), but it does not respect it when returning the validated data.

### Steps To Reproduce:
```
// $request->field1 = 100
$validated = $this->validate($request, [
    'field1' => [Rule::excludeIf(true), 'numeric']
]);

var_dump(isset($validated['field1']));
```

Expected output: **`FALSE`**
Actual output: **`TRUE`**

### Suggestion
I dug a bit deeper and see that `Laravel\Lumen\Routing\ProvidesConvenienceMethods` has its own method `extractInputFromRules`, which clearly does not factor in the exclude_if validation rule at all. My question is, why not just simplify the validate method to do exactly what Laravel does:

```
public function validate(Request $request, array $rules, array $messages = [], array $customAttributes = []) {
        return $this->getValidationFactory()->make($request->all(), $rules, $messages, $customAttributes)->validate();
}
```

I overrode the method in my application and all my tests are still passing. I have to assume it was done differently in Lumen for a reason, but I'm happy to submit a pull request if the above is acceptable.

driesvints:
> My question is, why not just simplify the validate method to do exactly what Laravel does:

Yes, feel free to send in a PR, thanks!

PR #1248, opened 2022-09-09
Change controller validate helper method to behave like Laravel (#1247)

MattApril:
#1247

### DIFF
--- a/src/Routing/ProvidesConvenienceMethods.php
+++ b/src/Routing/ProvidesConvenienceMethods.php
@@ -65,13 +65,14 @@ trait ProvidesConvenienceMethods
 
         try {
             $validated = $validator->validate();
+
             if (method_exists($this, 'extractInputFromRules')) {
-                // backwards compatability for deleted trait method
+                // Backwards compatability...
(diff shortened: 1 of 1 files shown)


driesvints:
Tests are failing here.

MattApril:
Whoops, I forgot that I overwrote my own exception handler. Looks like Lumen's default handler excepts the response to be set on the ValidationException. Fixed.

taylorotwell:
Would this break existing applications?

MattApril:
Yes, it would cause a breaking change if anyone overwrote the protected extractInputFromRules or throwValidationException methods. I just updated it to use those methods if they are defined.

halloei:
This broke my application. I'm not overwriting `extractInputFromRules()` and `throwValidationException()`, I'm calling them directly from my controller.

driesvints:
We're reverting this.

MattApril:
Hmm, that was an unfortunate oversight. Sorry about that.

I think this should be released as a breaking change in the next major version though. The current validation is not behaving as expected.

PR #1265, opened 2023-02-20
Allow PHPUnit v10

driesvints:

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,9 +17,8 @@ jobs:
       fail-fast: true
       matrix:
         php: [8.1, 8.2]
-        stability: [prefer-lowest, prefer-stable]
 
-    name: PHP ${{ matrix.php }} - ${{ matrix.stability }}
+    name: PHP ${{ matrix.php }}
 
     steps:
       - name: Checkout code
@@ -35,7 +34,7 @@ jobs:
           coverage: none
 
       - name: Install dependencies
-        run: composer update --${{ matrix.stability }} --prefer-dist --no-interaction --no-progress
+        run: composer update --prefer-stable --prefer-dist --no-interaction --no-progress
 
       - name: Execute tests
-        run: vendor/bin/phpunit --verbose
+        run: vendor/bin/phpunit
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yml
```diff
@@ -17,9 +17,8 @@ jobs:
       fail-fast: true
       matrix:
         php: [8.1, 8.2]
-        stability: [prefer-lowest, prefer-stable]
```
driesvints, line 20: I removed `prefer-lowest` from here to not have to bother about phpunit v9. I don't think it's necessary for Lumen anymore to check all of those lowest versions.

nunomaduro: If you don't want to bother with PHPUnit 9 just remove it from the `composer.json` file.

.github/workflows/tests.yml
```diff
@@ -17,9 +17,8 @@ jobs:
...
 
-    name: PHP ${{ matrix.php }} - ${{ matrix.stability }}
```
nunomaduro, line 22: Revert this.

.github/workflows/tests.yml
```diff
@@ -35,7 +34,7 @@ jobs:
           coverage: none
 
       - name: Install dependencies
-        run: composer update --${{ matrix.stability }} --prefer-dist --no-interaction --no-progress
```
nunomaduro, line 38: Revert this.

.gitignore
```diff
@@ -1,4 +1,5 @@
 /vendor
 composer.lock
 /phpunit.xml
+/.phpunit.cache
 .phpunit.result.cache
```
nunomaduro, line 5: Remove `.phpunit.result.cache`.

nunomaduro (changes requested):

.gitignore
driesvints: I think it's safer to leave this be because I already ran in the case where I accidentally almost committed this on my local Laravel 10 skeleton because I almost forgot to remove this file. Think a lot of contributors might run into the same issue.

nunomaduro (approved):

PR #1274, opened 2024-02-11
Update branding

caendesilva:
Updates the base branding

Additional changes we might want to consider:
- Update namespaces
- Update links and badges


caendesilva:
So sorry, this was not meant to be pushed to this fork. GitHub Desktop played a bamboozle on me, even after I explicably told it not to.

